PR #8, opened 2017-06-10
Add retry logic and format code with autopep8

chlela:


konvulse:
Frustrating.  My editor is supposed to format with Autopep8 as well.  It appears it has not been working.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+autopep8
```
konvulse: Not sure autopep8 is a requirement for the project.  Seems more like something that is part of the dev env.

chlela: it is recommended to keep consistent styling across the team, hence the requirement

konvulse: I agree regarding the styling.  But won't a user be required to install pep8 to use the library?

chlela: ohh I see what you mean. Yes yes, I agree with you

PR #9, opened 2017-06-15
Print response headers

konvulse:
Headers contain important information

### DIFF
--- a/conduce.py
+++ b/conduce.py
@@ -176,6 +176,8 @@ if __name__ == '__main__':
     try:
         result = args.func(args)
         if result:
+            if hasattr(result, 'headers'):
+                print result.headers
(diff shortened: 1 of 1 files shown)


clawconduce:
This seems like an improvement, but could be overwhelming.  Any thoughts on whitelisting/blacklisting specific headers?  "Location", "X-Error" could be whitelisted, or maybe cookies are blacklisted

konvulse:
@clawconduce I thought about a `-v` flag to print or not print the headers.  Maybe this is an indication that we should include critical information about the response in the body (in addition to the headers, in this case).

clawconduce:
It definitely seems worth considering.  Tagging @jrmConduce since he's designing the new API.

PR #10, opened 2017-06-15
teams and groups

konvulse:
Create, list, add users and list users

### DIFF
--- a/api.py
+++ b/api.py
@@ -113,6 +113,38 @@ def _make_get_request(uri, **kwargs):
     return response
 
 
+def create_team(team_name, **kwargs):
+    return make_post_request({'name': team_name}, 'team-create', **kwargs)
+
+
+def list_teams(**kwargs):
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -110,6 +110,37 @@ def _make_get_request(uri, **kwargs):
...
+def create_team(team_name, **kwargs):
+    return make_post_request({'name':team_name}, 'team-create', **kwargs)
+
+
+def list_teams(**kwargs):
+    return make_get_request('user/list-teams', **kwargs)
+
+
+def add_user_to_team(team_id, email_address, **kwargs):
+    return make_post_request({'email':email_address}, 'team/{}/invite'.format(team_id), **kwargs)
```
clawconduce: I believe this requires a wrapper `{"invitee": {email...}}`

konvulse: Oops, thanks.  I didn't test that one.

PR #14, opened 2017-06-20
Fixed exception handling

konvulse:


konvulse:
After merge I ran into issue with this code.  `HttpError` not found.

PR #15, opened 2017-06-20
Permissions API

konvulse:
Add bindings for setting permissions on conduce resources

### DIFF
--- a/api.py
+++ b/api.py
@@ -167,6 +167,59 @@ def list_group_members(team_id, group_id, **kwargs):
     return make_get_request('team/{}/group/{}/users'.format(team_id, group_id), **kwargs)
 
 
+def set_owner(team_id, resource_id, **kwargs):
+    return make_post_request({}, 'acl/{}/transfer-to-team/{}'.format(resource_id, team_id), **kwargs)
+
+
+def list_permissions(resource_id, **kwargs):
+    return make_get_request('acl/{}/view'.format(resource_id), **kwargs)
+
+
+def set_public_permissions(resource_id, read, write, share, **kwargs):
+    target_string = 'public'
+    return set_permissions(target_string, resource_id, read, write, share, **kwargs)
+
+
+def set_team_permissions(resource_id, read, write, share, **kwargs):
+    target_string = 'team'
+    return set_permissions(target_string, resource_id, read, write, share, **kwargs)
+
+
+def set_group_permissions(group_id, resource_id, read, write, share, **kwargs):
+    target_string = 'group/{}'.format(group_id)
+    return set_permissions(target_string, resource_id, read, write, share, **kwargs)
+
+
+def set_user_permissions(user_id, resource_id, read, write, share, **kwargs):
+    target_string = 'user/{}'.format(user_id)
+    return set_permissions(target_string, resource_id, read, write, share, **kwargs)
+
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -167,6 +167,44 @@ def list_group_members(team_id, group_id, **kwargs):
     return make_get_request('team/{}/group/{}/users'.format(team_id, group_id), **kwargs)
 
 
+def set_owner(team_id, resource_id, **kwargs):
```
clawconduce: I think this should mention team in the name somewhere so it is not confused with transferring to a user, which will probably get a function call one day

api.py
```diff
@@ -167,6 +167,44 @@ def list_group_members(team_id, group_id, **kwargs):
...
+    target_string = 'group/{}'.format(group_id)
+    return set_permissions(target_string, resource_id, read, write, share, **kwargs)
+
+
+def set_user_permissions(user_id, resource_id, read, write, share, **kwargs):
+    target_string = 'user/{}'.format(user_id)
+    return set_permissions(target_string, resource_id, read, write, share, **kwargs)
+
+
+def set_permissions(target_string, resource_id, read, write, share, **kwargs):
```
clawconduce: it might make sense to make this `_set_permissions`

api.py
```diff
@@ -167,6 +167,44 @@ def list_group_members(team_id, group_id, **kwargs):
...
 
+def set_owner(team_id, resource_id, **kwargs):
+    return make_post_request({}, 'acl/{}/transfer-to-team/{}'.format(resource_id, team_id), **kwargs)
+
+
+def list_permissions(resource_id, **kwargs):
+    return make_get_request('acl/{}/view'.format(resource_id), **kwargs)
+
+
+def set_public_permissions(resource_id, read, write, share, **kwargs):
```
clawconduce: I need to confirm this with @rclancey.

I am almost certain that on the grant acl endpoint only the changes what is passed in.  So one permissions, read/write/share, can be changed without affected the others.  I'd make these parameters kwargs that default to None, and if set to None, they do not get sent to the API & remain unchanged.

clawconduce: Unrelated so hopefully a different "conversation".  Should we have share permissions?  Seems like this feature is basically going to be unused.

api.py
```diff
@@ -167,6 +167,44 @@ def list_group_members(team_id, group_id, **kwargs):
...
+
+
+def set_permissions(target_string, resource_id, read, write, share, **kwargs):
+    permissions = {
+        'share': share,
+        'write': write,
+        'read': read,
+        'target': target_string
+    }
+    return make_post_request(permissions, 'acl/{}/view'.format(resource_id), **kwargs)
```
clawconduce: `view` --> `grant`

api.py
konvulse: Definitely a different conversation.  I reserve comment until that time.

konvulse: I realized this once I started testing.  See the changes.  Now `set_permissions` does a grant and revoke on every call, explicitly setting permissions.

api.py
konvulse: I see what you mean, but i'd rather `set_owner` be agnostic to whether it's setting the owner to a team or a user.  I may not have a choice here though.

PR #17, opened 2017-07-06
Begin to merge in OO version as a wrapper

clawconduce:
A long long time ago I said I would make an OO wrapper for the Python library, and the time has finally come, basically so the lib can survive the v2 transition.

This is not a complete wrapper by any means, but it's enough to run the sluice-to-banksy w/o major changes to the tool.

The goal is to eventually make this module a wrapper around the functional API.  The version in this PR is not all wrapper code.  But I think making the transition to a full wrapper makes sense after moving to the v2 api.

### DIFF
--- a/oo.py
+++ b/oo.py
@@ -4,13 +4,26 @@ import json
 import time
 import urlparse
 
+#this imports conduce.api
+import api as func_api
+
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 
 
-EXAMPLE_CONDUCE_API_KEY="tTPtOvbQrrSzgxR8uymnlpRzZIWEoEhTifJGLN1rZSs"
-VM_SERVER = "https://10.254.255.10"
 DEV_SERVER = "https://dev-app.conduce.com"
+
+#Resources
+USER_ASSET = "userassets"
+TEMPLATE = "templates"
+
+
+
+_LIST_KEYS = {
+    USER_ASSET: "asset",
+    TEMPLATE: "template",
+}
+
 
 class ConduceError(Exception):
     def __init__(self, url, resp, msg=None):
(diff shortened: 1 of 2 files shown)


oo.py
```diff
@@ -0,0 +1,276 @@
...
+
+class ConduceAPIClient(object):
+    #Simple Version history:
+    # 1.0 Version including all features to migrate using sluice-to-banksy & run a RT daemon
+    # 2.0: Use v2 datasets for everything
+    # 2.1: Make create dataset sage function & ensure_sage_dataset function
+    # 3.0: Integrate with functional api
+    __version = 3.0
+
+    def __init__(self, host="https://app.conduce.com", api_key=None, email=None, password=None, insecure=True):
```
rclancey: default for insecure should be false.

oo.py
```diff
@@ -0,0 +1,276 @@
...
+            email = None
+            password = None
+        #TODO: Make sure API has valid characters letters, numbers & - & _
+        self._api_key = api_key
+        self._email = email
+        self._password = password
+        self._cookies = None
+        self._verify = not insecure
+        info = urlparse.urlparse(host)
+        scheme = info.scheme if info.scheme else "https"
```
rclancey: Is this ever not https?  might simplify to make host be just the hostname, not the url, and always use https

oo.py
```diff
@@ -0,0 +1,276 @@
...
+        headers = {'Content-Type': 'application/json'}
+        req = requests.post('%s/conduce/api/v1/user/login' % self._host, data=login_data, headers=headers, verify=self._verify)
+        if not req.ok:
+            raise ConduceError(req.url, req)
+        self._cookies = req.cookies
+        #HACK!!!
+        req = self.api_get("/apikeys/list")
+        apikeys = req.json()
+        if len(apikeys) == 0:
+            raise NotImplementedError("Create an API key under the {} user to use the OO API".format(self._email))
```
rclancey: Any reason to not just create one here?

rclancey: Also, if we can list api keys, we can certainly do the rest of the operations without an api key, can't we?

oo.py
```diff
@@ -0,0 +1,276 @@
...
+    # Impedence matching the functional API
+    #
+    def _get_creds(self):
+        if self._api_key:
+            return {"api_key": self._api_key, "host": self._hostname}
+        else:
+            raise NotImplementedError("The object oriented API does not support the email/password login yet")
+
+    def _check_resp(self, resp):
+        if not resp.ok:
```
rclancey: does resp.ok cover all of the 20x responses (like 202 for data ingest apis)?

oo.py
```diff
@@ -0,0 +1,276 @@
...
+
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
+
+
+DEV_SERVER = "https://dev-app.conduce.com"
+
+#Resources
+USER_ASSET = "userassets"
+TEMPLATE = "templates"
```
rclancey: orchestrations and substrates should work with the resources API, FYI

konvulse:
This is probably out of date.

clawconduce:
It def. is, I will give this another pass to update so it works with the v2 api.

I will probably do an additional pass to start getting everything to converge more - so breaking the interface in hopefully minimal ways & updating sluice-to-banksy.

PR #18, opened 2017-07-08
Updated readme for module installation

konvulse:
Remove the references to running the command line utility from source
and provide instructions for installing with pip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,23 +6,28 @@ This library implements the most commonly used Conduce API endpoints.  It can be
 
 # Installation
 
-If you want this library available to import for various functions and want command line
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,10 +79,10 @@ A basic configuration looks like:
 default-host: dev-app.conduce.com
 default-user: enterprise_test@conduce.com
 dhl-dev@conduce.com:
-  api-keys:
```
clawconduce: was this intended?  It seems like the map of env to api key should be under the api-keys section

konvulse: Thanks, that was not intentional.

PR #19, opened 2017-07-08
Added acl library

konvulse:
Move acl functions out of dhl-pilot.

This requires documentation on the format of the permissions and teams object structures.

### DIFF
--- a/acl.py
+++ b/acl.py
@@ -1,0 +1,114 @@
+import json
+import api
+
+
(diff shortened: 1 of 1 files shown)


konvulse:
Did you notice that the acl functions are in `acl.py` and not `api.py`?  As such they are not part of the `conduce.api` module, but rather `conduce.acl`.  I thought that was sufficient.

PR #23, opened 2017-07-12
Check for and create accounts

konvulse:


acl.py
```diff
@@ -76,11 +76,23 @@ def build_groups(groups, team_id, **kwargs):
...
-            user_id = get_user_id(user['email'], team_members)
-            if user_id is None:
-                api.add_user_to_team(team_id, user['email'], **kwargs)
-            if user_id is not None and not user_in_group(user_id, group_members):
-                api.add_user_to_group(team_id, group_id, user_id, **kwargs)
+            if api.account_exists(user['email'], **kwargs):
+                user_id = get_user_id(user['email'], team_members)
+                if user_id is None:
+                    print 'Adding {} to team'.format(user['email'])
+                    api.add_user_to_team(team_id, user['email'], **kwargs)
```
clawconduce: what happens if they are already on the team?

chlela: user_id would not be None?

api.py
```diff
@@ -693,3 +693,18 @@ def remove_asset(**kwargs):
...
+        return False
+    except requests.exceptions.HTTPError as e:
+        if e.response.status_code == 409:
+            return True
+        else:
+            raise e
+
+
+def create_account(name, email, **kwargs):
+    return make_post_request({'email': email, 'name': name}, 'admin/create-user', **kwargs)
```
chlela: no error handling? or retrying?

api.py
```diff
@@ -693,3 +693,18 @@ def remove_asset(**kwargs):
 
 def _remove_asset(asset_id, **kwargs):
     return make_post_request({}, 'userassets/delete/{}'.format(asset_id), **kwargs)
+
+
+def account_exists(email, **kwargs):
+    try:
+        make_post_request({'email': email}, 'user/public/email-available', **kwargs)
```
chlela: Something feels weird with this approach.  I'm assuming you're using post request to hide email address from a sniffer.  Now the URL states 'email-available' and returns a conflict 409 error.  Isn't 404 not found more appropriate here?  Which answers the question for URL 'email-available', success means yes, exception means no.

acl.py
konvulse: `get_user_id` looks at the list provided and returns the ID for the matching email address.  This informs as to whether the user is part of the team.  So if the user is on the team, `get_user_id` _should_ return the ID, otherwise `None`.

api.py
konvulse: I thought that was handled my `_make_post_request` and `_make_get_request`.

konvulse: The response is returned to the user and exceptions bubble up to the user for error handling.

acl.py
konvulse: There is a bug here though. If the user is not on the team they are not added to the group, since `user_id` is None, and not updated after the user is added to the team.

konvulse:
The new commit handles the case I described where users are not added to a group if they were added to the team.

PR #24, opened 2017-07-21
Add clear dataset command

konvulse:
Added `/datasets/clear/{id}` API method and did some CLI arg parse cleanup.


konvulse:
GitHub auto-populated the title with the wrong commit message.  Should have been, "Added clear-dataset command."

PR #25, opened 2017-08-07
Remove email addresses from examples

konvulse:


elliottconduce:
@konvulse thank you for making this change.

PR #26, opened 2017-08-08
Added ingest data command

konvulse:
Update create-dataset and added ingest-data subcommand.  Also updated
ingest to populate endtime with timestamp if no endtime was provided in
the data.

### DIFF
--- a/api.py
+++ b/api.py
@@ -219,13 +219,6 @@ def set_permissions(target_string, resource_id, read, write, share, **kwargs):
 def create_dataset(dataset_name, **kwargs):
     response = make_post_request(
         {'name': dataset_name}, 'datasets/create', **kwargs)
-    response_dict = json.loads(response.content)
-    if 'json' in kwargs and kwargs['json']:
-        ingest_entities(response_dict['dataset'], json.load(
-            open(kwargs['json'])), **kwargs)
-    elif 'csv' in kwargs and kwargs['csv']:
-        ingest_entities(response_dict['dataset'], util.csv_to_entities(
-            kwargs['csv']), **kwargs)
 
     return response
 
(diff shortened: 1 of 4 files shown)


konvulse:
There is general cleanup here as well.  I removed some unnecessary `raise_for_status` calls.  `raise_for_status` is called by `_make_get_request` and `_make_post_request`.  It is not necessary to call `raise_for_status` anywhere else.

api.py
```diff
@@ -219,13 +219,6 @@ def set_permissions(target_string, resource_id, read, write, share, **kwargs):
 def create_dataset(dataset_name, **kwargs):
     response = make_post_request(
         {'name': dataset_name}, 'datasets/create', **kwargs)
-    response_dict = json.loads(response.content)
```
konvulse: This is not part of creating a dataset.  It belongs in `util`.

api.py
```diff
@@ -276,20 +269,21 @@ def ingest_entities(dataset_id, data, **kwargs):
...
         job_id = response.headers['location']
         response = wait_for_job(job_id, **kwargs)
+
     return response
 
 
 def _clear_dataset(dataset_id, **kwargs):
     response = make_post_request(
         None, 'datasets/clear/{}'.format(dataset_id), **kwargs)
-    response.raise_for_status()
```
konvulse: Please double check me, but these are unnecessary checks.

conduce.py
```diff
@@ -17,7 +18,21 @@ def list_datasets(args):
 
 
 def create_dataset(args):
-    return api.create_dataset(args.name, **vars(args))
```
konvulse: User can still create a dataset with data.

conduce.py
```diff
@@ -17,7 +18,21 @@ def list_datasets(args):
...
+    response_dict = json.loads(response.content)
+
+    if args.json or args.csv:
+        print json.dumps(response_dict, indent=2)
+        response = util.ingest_file(response_dict['dataset'], **vars(args))
+
+    return response
+
+
+def ingest_data(args):
```
konvulse: User can ingest data without creating a dataset.

util.py
```diff
@@ -218,7 +218,10 @@ def get_default(field):
 
 def get_field_value(raw_entity, key_map, field):
     if not key_map[field]['key']:
-        return get_default(field)
+        if field == 'endtime_ms':
+            return get_field_value(raw_entity, key_map, 'timestamp_ms')
```
konvulse: Default endtime to timestamp.

PR #29, opened 2017-08-17
Added Sphinx to requirements.txt

konvulse:


clawconduce:
I haven't followed these, but maybe this could go into setup.py under a special field.  I have seen `build_requires` and maybe `install_requires` and they mean different things.  This shouldn't be required to install, but may be required to build.

I didn't look hard, but I found `test_requires` also exists: https://stackoverflow.com/questions/15834628/how-to-specify-explicit-python-packaging-dependencies-in-setup-py

PR #31, opened 2017-08-29
Apply user specified timezone to parsed datetime

konvulse:
Allows the timezone in the string to be replaced by a user specified timezone.  Primarily to be used in cases where the timezone is not specified in the string and the timestamp is known to be in local time.


konvulse:
I need to keep making progress.  A review on this would still be appreciated.

PR #34, opened 2017-09-06
v2 Resource API support

konvulse:
Removed endpoints for orchestration, userasset, substrate, and template.  Use resource API instead.  Datasets, substrates, orchestrations, assets, and lens templates are now created and destroyed with the resource API.

Raw resources are returned in their raw format.  All others are returned as a list or dictionary.  

Convenience methods are provided where the REST API takes a `type` parameter (create, find, remove).

There is no implementation of the PATCH API at this time.

### DIFF
--- a/conduce.py
+++ b/conduce.py
@@ -16,6 +16,23 @@ def list_from_args(args):
     return api.list_resources(object_to_list.upper().rstrip('S'), **vars(args))
 
 
+def find_resource(args):
+    if args.type is not None:
+        resource_type = args.type
+
+        if resource_type.lower() == "lenses" or resource_type.lower() == "templates":
+            resource_type = "LENS_TEMPLATE"
+
+        args.type = resource_type.upper().rstrip('S')
+
+    resources = api.find_resource(**vars(args))
+    if args.content == 'full':
+        for resource in resources:
+            resource['content'] = json.loads(resource['content'])
+
+    return resources
+
+
 def list_datasets(args):
     return api.list_datasets(**vars(args))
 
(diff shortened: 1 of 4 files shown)


jrmConduce:
It is perhaps beyond the scope of this PR, but it might be nice to add support for the new v2 data API as well: /data/{datasetId}/transactions

That plus the resource API is what we need to implement a complete import/export outside of the backend.  It would be good to be able to use the python API to do that.

konvulse:
It is out-of-scope.  Data will come next.

conduce.py
```diff
@@ -10,7 +10,27 @@ def list_from_args(args):
...
+        object_to_list = "LENS_TEMPLATE"
+
+    return api.list_resources(object_to_list.upper().rstrip('S'), **vars(args))
+
+
+def find_resource(args):
+    if args.type is not None:
+        resource_type = args.type
+
+        if resource_type.lower() == "lenses" or resource_type.lower() == "templates":
```
clawconduce: nitpick: we may instead make this raise an exception that returns the valid list of resource types

konvulse: If there is an API to get that list, maybe.  Otherwise we don't want to maintain that here.

Besides we do raise in that case.  We display the error message returned by the server

```
Kevins-MacBook-Pro:conduce-python-api kvl$ conduce-api list otherthings
400 Client Error: Bad Request for url: https://dev-app.conduce.com/conduce/api/v2/resources/searches
Could not read json request: cannot unmarshal input

```

:/

PR #35, opened 2017-09-08
Added patch support, kindof

konvulse:


api.py
```diff
@@ -912,6 +912,39 @@ def create_asset(name, content, mime_type, **kwargs):
...
+            content = base64.b64encode(content)
+
+    return _modify_resource_content(resource_id, content, **kwargs)
+
+
+def _modify_resource_content(resource_id, content, **kwargs):
+    # payload = [{
+    #    'path': '/content',
+    #    'value': content,
+    #    'op': 'add'
```
jrmConduce: Note that the fix for this has been merged into dev.  So you can now correctly use replace.

konvulse: Not until it gets to dhl-prd

konvulse: Got it.

PR #36, opened 2017-09-08
v2 api cleanup

konvulse:
This PR follows the patch PR and does some general cleanup.

### DIFF
--- a/api.py
+++ b/api.py
@@ -402,7 +402,7 @@ def clear_dataset(**kwargs):
         datasets = list_datasets(**kwargs)
         to_clear = []
         for dataset in datasets:
-            if dataset.get('name', 0) == kwargs['name'] or (kwargs['regex'] and re.match(kwargs['regex'], dataset.get('name', 0))):
+            if dataset.get('name') is not None and (dataset['name'] == kwargs['name'] or (kwargs['regex'] and re.match(kwargs['regex'], dataset['name']))):
                 to_clear.append(dataset)
         if len(to_clear) == 1:
             _clear_dataset(to_clear[0]['id'], **kwargs)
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -405,7 +402,7 @@ def clear_dataset(**kwargs):
         datasets = list_datasets(**kwargs)
         to_clear = []
         for dataset in datasets:
-            if dataset['name'] == kwargs['name'] or (kwargs['regex'] and re.match(kwargs['regex'], dataset['name'])):
+            if dataset.get('name', 0) == kwargs['name'] or (kwargs['regex'] and re.match(kwargs['regex'], dataset.get('name', 0))):
```
clawconduce: I believe this is aiming for robustness, but it will fail on regex:

```
> re.match(".", 0)
TypeError: expected string or buffer
```
What about checking dataset up front?

`if dataset.get('name') and (dataset["name"] == kwargs['name'] or (kwargs['regex'] and re.match(kwargs['regex'], dataset['name']))):`

clawconduce: I'll make a PR for this

conduce.py
```diff
@@ -26,9 +27,13 @@ def find_resource(args):
...
     resources = api.find_resource(**vars(args))
-    if args.content == 'full':
-        for resource in resources:
-            resource['content'] = json.loads(resource['content'])
+    if args.decode:
+        if args.content == 'full':
+            for resource in resources:
+                if resource.get('mime', 'invalid-mime') == 'application/json':
+                    resource['content'] = json.loads(resource['content'])
+                elif 'text' not in resource:
```
clawconduce: `not in resource.get("mime")`?  text isn't a field on resources

clawconduce:
@konvulse I put some fixes in, check out the 2 commits, with those this PR LGTM

PR #42, opened 2017-10-25
Feature/dhl 294 update team users on adding user

clawconduce:


clawconduce:
Yes, I thought that would be less robust - we might change the information/format that is saved in that variable which might matter if we expand the function

PR #43, opened 2017-10-27
Documentation

konvulse:
Added ingest documentation.

I think it is time to baseline this work in order to move forward with API changes that sync to the documentation.


doc/source/conduce-entities.rst
```diff
@@ -0,0 +1,159 @@
...
+
+The data in this table represents the present location of 5 shipments.  Each shipment has an ID, a shipping method, a date when it was last updated and a location in geographic coordinates.  Additionally the current state and value of the shipment is documented.  A Conduce entity is able to capture and represent all of this information.  These data records will be ingested as the first samples of 5 unique entities of the ``shipments`` dataset.
+
+An entity sample requires four fields:
+ **id**
+     a unique string that represents a specific object that exists.
+ **kind**
+     the type or category of the object.
+ **time**
+     A :py:class:`datetime.datetime` at which the entity state described by this sample is valid.
```
clawconduce: might be worth mentioning more info here since it's easy to mess this up.

Like, if no timezone is present, use UTC/localtime.  Also, mention `datetime.now()` has no timezone but returns local time, so users should use `datetime.utcnow()` which also has no timezone, but returns UTC time.  Maybe this info doesn't go here, but goes somewhere nearby

doc/source/conduce-entities.rst
```diff
@@ -0,0 +1,159 @@
...
+
+And one of:
+ **point**
+     A coordinate pair that defines a location in a 2D cartisian coordinate system.
+ **path**
+     A list of coordinates that define a series of connected line segments in a 2D cartisian coordinate system.
+ **polygon**
+     A list of coordinates that define a closed shape in a 2D cartisian coordinate system.
+
+Each can be specified in ``(x,y)`` or ``(latitude,longitude)``. If coordinate types are mixed or multiple coordinate types are specified an error will result.  If more than one location is defined, only one will be used.  Precidence is undefined.
```
clawconduce: super nitpick: This might be a little confusing since x,y & lat,lon have special formatting & look like they could translate easily to code.  So it sort of looks a user can use tuples.  Also, x, y are real key values, but `latitude` & `longitude` are not.

doc/source/conduce-entities.rst
```diff
@@ -0,0 +1,159 @@
...
+        "time": dateutil.parser.parse('2017-10-24T10:22:09+00:00'),
+        "point": {
+            "lon": -91.571045,
+            "lat": 38.022131
+        },
+        "Value": 102325.26,
+        "Status": "delivered" 
+    }
+
+In the example above you see that the ISO-8601 date time string is converted to a :py:class:`datetime.datetime` using :py:func:`dateutil.parser.parse`.  This integer represents the number of milliseconds that have accumulated since epoch (``1970-01-01T00:00:00+00:00``)
```
clawconduce: Nix "This integer represents the number of milliseconds that have accumulated since epoch (``1970-01-01T00:00:00+00:00``)"

doc/source/conduce-entities.rst
```diff
@@ -0,0 +1,159 @@
...
+     - 25.029996
+     - -134.318848
+     - $1,431,873,345.01
+     - transit
+
+The data in this table represents the present location of 5 shipments.  Each shipment has an ID, a shipping method, a date when it was last updated and a location in geographic coordinates.  Additionally the current state and value of the shipment is documented.  A Conduce entity is able to capture and represent all of this information.  These data records will be ingested as the first samples of 5 unique entities of the ``shipments`` dataset.
+
+An entity sample requires four fields:
+ **id**
+     a unique string that represents a specific object that exists.
```
clawconduce: In general, I think it should be clearer that an entity = ID.  So all the samples for a given entity point to the same ID.

Right here, I'd say "a unique string that represents a specific ~object~ entity that exists".

doc/source/conduce-entities.rst
```diff
@@ -0,0 +1,159 @@
...
+
+.. note:: WIP
+
+Conduce uses a concept called entity to define objects that exist in space and time.  An entity can exist for an instant or across a time span.  It can describe a point in space, a path (sequence of points), or a region (polygon).  An entity is defined by a series of samples that describe it's location in space over a period of time.  An entity also has attributes that describe it's characteristics. An entity's attributes can change over time.
+
+-----------------
+Creating entities
+-----------------
+
+An entity is a series of samples that describe the state of and object over a period of time.  A sample is the representation of an entity's momentary state.  Samples are created from user data.  The data should describe people or things that exist in some location during some time span.  Data can be ingested from static files with thousands of records or real-time streams that provide state updates in real-time.  As an example we'll ingest the following real-time shipment data into an existing dataset named ``shipments``:
```
clawconduce: 2 suggestion for this paragraph:

1) Somewhere in here I'd call out how ID points to an entity, and a sample is info for a specific time.  For example, change the 4th sentence to: "Sample data should identify a person or thing - an entity - and describe its location and state at a given moment in time."

2) When I read this I think, "how do I create an entity?".  To me this sounds like a user has to create an entity & then add samples.  It should be clear that only samples need to be ingested.

api.py
```diff
@@ -64,6 +64,28 @@ def list_templates(**kwargs):
...
+    """
+    Wait for a job to complete.
+
+    Block execution until the completion of an asynchronous job.
+
+    Parameters
+    ----------
+    job_id : string
+        The UUID that identifies the job to query.  The job ID is returned in the response header of the request that starts the job.
+    **kwargs:
```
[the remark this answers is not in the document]
konvulse: What makes `ConduceCreds` different from `kwargs`?

doc/source/conduce-entities.rst
konvulse: I don't want to take responsibility for the datetime interface.  That is why I link to it.

doc/source/conduce-entities.rst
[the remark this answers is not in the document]
konvulse: That's not a nit.  Spelling counts.

konvulse: True.

doc/source/conduce-entities.rst
konvulse: Yes, the overall language may need to change.  This document is really about samples now.  When we started it was about entities.  In the first sentence we state that, "an entity is a series of samples".

api.py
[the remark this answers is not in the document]
clawconduce: `ConduceCreds` would be an object with only the parameters required to get onto a specific server, like host & api-key, or host & email.  Flags like `wipe`, `create-accounts`, `orch` and `acl` couldn't hitch a ride

doc/source/conduce-entities.rst
konvulse: Oops

doc/source/conduce-entities.rst
clawconduce: That's fair.

I would still mention something like, "It is recommend users always include specify a timezone on the datetime object.  If no timezone is specified, we (assume UTC|assume localtime|throw an exception)"

doc/source/conduce-entities.rst
konvulse: I am trying to relate user data to entities/samples here.  That is why I say "the data should describe..." and don't mention entities.  I also don't think IDs are relevant at this more conceptual level.

We are implicitly declaring that entities are unique when we say things like "the state of an object over a period".

api.py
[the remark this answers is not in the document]
konvulse: What's the harm if they do?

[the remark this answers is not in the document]
clawconduce: I have a few reasons.  As having used docs for python libs that use kwargs, it's prone to error and leads to inconsistency over time.

When I use these libs it's frustrating as a user because kwargs is just a vague bucket.  I have to go hunting for info. Oftentimes I have variables I want to pass in, like login info, and just want to see parameter names since I can figure out what they mean (I want to delete something, and I see remove, that must be what I want to use - or is it email? user? account? - or is it host? url? env? server?).  I also want a low friction way to see all the parameters I can use to change behaviour.  For example `help(json.dumps)` shows `dumps(obj, skipkeys=False, ensure_ascii=True, check_circular=True, allow_nan=True, cls=None, indent=None, separators=None, encoding='utf-8', default=None, sort_keys=False, **kw)` as the signature.  The names are descriptive and it makes it really easy to see what sort of knobs I can tweak.  There is a `**kw` in there, and finding more info on that is not something I want to do.  Reading the docstring to find out is enough work, I don't want to go looking in more places.

I also not-uncommly find errors with `**kwargs` doc strings too.  Code changes and the docs get out of date.  Someone updates the kwargs that get used, but forgets to update the docs.  If kwargs are explicitly specified in the function signature that's always up to date.

Additionally, it is easy to mess things up.  I can have a typo in a `**kwarg` key and then all bets are off - sometimes `**kwargs` is written to use defaults, sometimes it crashes.  If I a kwarg is explicitly specified, and I have a typo, I will get a `TypeError got an unexpected keyword argument 'mytypo'`.  Along the same lines, if I pass in a blob of kwargs into a function, I might get unlucky & have a collision - the `orch` flag is True b/c I want to load the orch yaml config, but maybe a permissions function will share an orch if its id is given under `kwargs[orch]` & then I hit an error because I tried to share an orch with an ID of `True`.

From a maintainer perspective, I'd say the previous point is a big one & encapsulation is important - it's easy to see what info is used where.  It'd be easy to make observations like, this command line args go into the credentials which are passed around to anything that uses the API.  The create-accounts flag is only used when building the teams, etc.

api.py
```diff
@@ -368,6 +389,32 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
 
 
 def ingest_entities(dataset_id, data, **kwargs):
+    """
+    Upload an :ref:`entity set <conduce-entities>` to the Conduce datastore.
```
jrmConduce: Do we want to keep using the term entity set here and immediately below?

You're now referring to it as a list of entities elsewhere

doc/source/conduce-entities.rst
```diff
@@ -0,0 +1,159 @@
...
+
+More :ref:`example sample definitions <entity-sample-definitions>`
+
+------------------
+Ingesting entities
+------------------
+
+In order to ingest our source data we must first convert each record into an entity sample as described above.  Once converted, the samples are added to a list.  The list may contain samples for multiple entities.  Once we have created our sample list we call :py:func:`api.ingest_entities`::
+
+    ingest_entities(dataset_id, sample_list, host=app.conduce.com, api-key=00000000-0000-0000-0000-000000000000)
```
jrmConduce: Is this correct?  Wasn't the intent to have `ingest_samples` vs `ingest_entities`?

doc/source/entity-sample-definitions.rst
```diff
@@ -0,0 +1,194 @@
...
+        "id": 1,
+        "kind": "truck",
+        "time": datetime.datetime.now()
+        "point": {
+            "lon": -91.571045,
+            "lat": 38.022131
+        },
+    }
+
+A geographic sample could also be represented with a point sample.::
```
jrmConduce: While it is implied given the example, it may be worth explicitly stating that x maps to longitude and y maps to latitude

doc/source/conduce-entities.rst
konvulse: That change will come with the next PR.  `ingest_samples` does not exist at the time of this writing.

PR #44, opened 2017-10-27
SKW-1981: Add ingest_samples API

konvulse:
This PR adds the `ingest_samples` API.  This API accepts sample lists as defined in sphinx docs.

To build the sphinx docs run `make html` in the `docs` directory of the repo.  You should also pip install requirements.txt.

### DIFF
--- a/api.py
+++ b/api.py
@@ -418,7 +418,7 @@ def convert_samples_to_entity_set(sample_list):
         coordinates = []
         if 'point' in sample:
             coordinates = [convert_coordinates(sample['point'])]
-        if 'coordinates' in sample:
+        if 'path' in sample:
             if len(coordinates) > 0:
                 raise KeyError('A sample may only contain one of point, path or polygon', sample)
             if len(sample['path']) < 2:
@@ -428,7 +428,7 @@ def convert_samples_to_entity_set(sample_list):
         if 'polygon' in sample:
             if len(coordinates) > 0:
                 raise KeyError('A sample may only contain one of point, path or polygon', sample)
-            if len(sample['path']) < 3:
+            if len(sample['polygon']) < 3:
                 raise KeyError('polygons must have at least three points')
             for point in sample['polygon']:
                 coordinates.append(convert_coordinates(point))
--- a/api.py
+++ b/api.py
@@ -409,6 +409,28 @@ def convert_coordinates(point):
     return coord
 
 
+def convert_geometries(sample):
+    coordinates = []
+    if 'point' in sample:
+        coordinates = [convert_coordinates(sample['point'])]
+    if 'path' in sample:
+        if len(coordinates) > 0:
+            raise KeyError('A sample may only contain one of point, path or polygon', sample)
+        if len(sample['path']) < 2:
+            raise KeyError('paths must have at least two points')
+        for point in sample['path']:
+            coordinates.append(convert_coordinates(point))
+    if 'polygon' in sample:
+        if len(coordinates) > 0:
+            raise KeyError('A sample may only contain one of point, path or polygon', sample)
+        if len(sample['polygon']) < 3:
+            raise KeyError('polygons must have at least three points')
+        for point in sample['polygon']:
+            coordinates.append(convert_coordinates(point))
+
+    return coordinates
+
+
 def convert_samples_to_entity_set(sample_list):
     conduce_keys = ['id', 'kind', 'time', 'point', 'path', 'polygon']
     entities = []
@@ -430,24 +452,7 @@ def convert_samples_to_entity_set(sample_list):
         if not isinstance(sample['time'], datetime):
             raise TypeError('Error processing sample at index {}. Time must be a datetime object.'.format(idx), sample['time'])
 
-        coordinates = []
-        if 'point' in sample:
-            coordinates = [convert_coordinates(sample['point'])]
-        if 'path' in sample:
-            if len(coordinates) > 0:
-                raise KeyError('A sample may only contain one of point, path or polygon', sample)
-            if len(sample['path']) < 2:
-                raise KeyError('paths must have at least two points')
-            for point in sample['path']:
-                coordinates.append(convert_coordinates(point))
-        if 'polygon' in sample:
-            if len(coordinates) > 0:
-                raise KeyError('A sample may only contain one of point, path or polygon', sample)
-            if len(sample['polygon']) < 3:
-                raise KeyError('polygons must have at least three points')
-            for point in sample['polygon']:
-                coordinates.append(convert_coordinates(point))
-
+        coordinates = convert_geometries(sample)
         if coordinates == []:
             raise ValueError('Error processing sample at index {}.  Samples must define a location (point, path, or polygon)'.format(idx), sample)
 
--- a/api.py
+++ b/api.py
@@ -8,6 +8,7 @@ import re
 import urlparse
 import base64
 import warnings
+from datetime import datetime
 
 from retrying import retry
 
@@ -96,7 +97,6 @@ def wait_for_job(job_id, **kwargs):
             if 'response' in msg:
                 return response
         else:
-            print response.status_code, response.text
             return response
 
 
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -367,7 +389,130 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
     return ingest_entities(dataset_id, {'entities': [entity]}, **kwargs)
 
 
+def convert_coordinates(point):
+    if len(point) > 2:
+        raise KeyError('A point should only have two dimensions', point)
```
konvulse: I used `KeyError` a lot.  Basically, any time the data structure is not in compliance with the documentation.

`KeyError` is documented as:
>Raised when a mapping (dictionary) key is not found in the set of existing keys.

api.py
```diff
@@ -367,7 +389,130 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+
+    return coord
+
+
+def convert_samples_to_entity_set(sample_list):
+    conduce_keys = ['id', 'kind', 'time', 'point', 'path', 'polygon']
+    entities = []
+    for sample in sample_list:
+        if not isinstance(datetime, sample['time']):
+            raise TypeError('time must be a datetime object')
```
konvulse: `TypeError` seems right here

api.py
```diff
@@ -367,7 +389,133 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+                raise KeyError('A sample may only contain one of point, path or polygon', sample)
+            if len(sample['path']) < 3:
+                raise KeyError('polygons must have at least three points')
+            for point in sample['polygon']:
+                path.append(convert_coordinates(point))
+
+        attribute_keys = [key for key in sample.keys() if key not in conduce_keys]
+        attributes = util.get_attributes(attribute_keys, sample)
+
+        attributes['_kind'] = sample['kind']
```
konvulse: What do we think of this attribute `_kind`?

api.py
```diff
@@ -367,7 +389,133 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+        if 'polygon' in sample:
+            if len(path) > 0:
+                raise KeyError('A sample may only contain one of point, path or polygon', sample)
+            if len(sample['path']) < 3:
+                raise KeyError('polygons must have at least three points')
+            for point in sample['polygon']:
+                path.append(convert_coordinates(point))
+
+        attribute_keys = [key for key in sample.keys() if key not in conduce_keys]
+        attributes = util.get_attributes(attribute_keys, sample)
```
konvulse: I reused `util.get_attributes` but it does conversions from float to int.  Take a look and see what you think.  I don't think it would be a problem to remove the conversion to int and leave all attributes in floating point.

doc/source/entity-sample-definitions.rst
```diff
@@ -180,7 +180,7 @@ Attributes are used to define characteristics of entities not encompassed by the
     {
         "id": 1,
         "kind": "ball",
-        "time": 1508840529000,
+        "time": dateutil.parser.parse(dateutil.parser.parse(1508840529000)),
```
clawconduce: `parse(parse)`?

api.py
clawconduce: I'd go for "kind" and put attrs in their own nested map

api.py
clawconduce: +1

doc/source/entity-sample-definitions.rst
```diff
@@ -180,7 +180,7 @@ Attributes are used to define characteristics of entities not encompassed by the
     {
         "id": 1,
         "kind": "ball",
-        "time": dateutil.parser.parse(dateutil.parser.parse(1508840529000)),
+        "time": dateutil.parser.parse(dateutil.parser.parse(2017-10-27T10:23:32+05:00)),
```
clawconduce: why `parse(parse())`?

clawconduce:
I think this needs to be diffed against the open PR for documentation before it can be properly reviewed - I was going to go by commit, but I see my comments are outdated even though none of the later commits (per github) touch the files I commented on

api.py
```diff
@@ -367,7 +388,134 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+def convert_samples_to_entity_set(sample_list):
+    conduce_keys = ['id', 'kind', 'time', 'point', 'path', 'polygon']
+    entities = []
+    for sample in sample_list:
+        if not isinstance(sample['time'], datetime):
+            raise TypeError('time must be a datetime object', sample['time'])
+        coordinates = []
+        if 'point' in sample:
+            coordinates = [convert_coordinates(sample['point'])]
+        if 'coordinates' in sample:
```
chlela: did you mean if 'path' in sample?

api.py
```diff
@@ -367,7 +388,134 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+            if len(coordinates) > 0:
+                raise KeyError('A sample may only contain one of point, path or polygon', sample)
+            if len(sample['path']) < 2:
+                raise KeyError('paths must have at least two points')
+            for point in sample['path']:
+                coordinates.append(convert_coordinates(point))
+        if 'polygon' in sample:
+            if len(coordinates) > 0:
+                raise KeyError('A sample may only contain one of point, path or polygon', sample)
+            if len(sample['path']) < 3:
```
chlela: did you mean if sample['polygon']

util.py
```diff
@@ -9,6 +9,15 @@
 import pytz
 
 
+def get_dataset_id(dataset_name, **kwargs):
+    datasets = api.list_datasets(**kwargs)
+    for dataset in datasets:
```
chlela: nit: can this for loop be condensed into python [d for d in ...] pattern?

util.py
```diff
@@ -27,6 +36,12 @@ def format_mac_address(mac_address):
     return re.sub('[:-]', '', mac_address).upper().strip()
 
 
+def datetime_to_timestamp_ms(dt):
```
chlela: should we move datetime helpers to conduce-python-api from dhl-helpers so they can be used in dhl-pilot, dhl-ingest and conduce-python-api?

konvulse:
@clawconduce -- This PR is for the new `ingest_samples` API.  Comments on the documentation should be made in #43

api.py
konvulse: No this new API method uses new data structures outlined in the new sphinx docs.  See PR #43.  I'll send you an invite to hosted sphinx documentation.

konvulse: Thanks I had some inconsistent naming.  Looks like I didn't get it all resolved.

api.py
konvulse: Yes

util.py
konvulse: Good catch.  I didn't know that was in both places.  I think it makes sense for the timestamp helpers to be in conduce-python-api.

We might just be able to remove them from dhl-helpers.

util.py
konvulse: It could, but I didn't touch this code.  I just reorganized the file.

chlela: 👍

util.py
```diff
@@ -342,21 +354,16 @@ def dict_to_entities(raw_entities, **kwargs):
...
-            return dataset['id']
-
-    return None
+def csv_to_entities(infile, outfile=None, toStdout=False):
+    return dict_to_entities(csv_to_json(infile))
 
 
 def ingest_json(dataset_id, json_file, **kwargs):
-    api.ingest_entities(dataset_id, json.load(open(json_file)), **kwargs)
+    api._ingest_entity_set(dataset_id, dict_to_entities(json.load(open(json_file))), **kwargs)
```
chlela: prefixing a function with underscore is a protocol for private function that shouldn't be accessible from other libraries. I would recommend that we avoid this approach or rename the function if it is public facing.

or we can move _ingest_entity_set to util file and have util.ingest_csv, util.ingest_json, util.ingest_samples live next to each other and be used by other libraries

I like the second approach better.  But also have concerns about second approach and I feel like those three methods should live in their own class referenced by other libraries.  When I think of utilities class, I view it as an independent set of methods that either format things, calculate things or transform things.  Having util depend on api library is questionable. Feels like they belong in their own library, `api_utils` maybe?

api.py
```diff
@@ -367,7 +388,134 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
     return ingest_entities(dataset_id, {'entities': [entity]}, **kwargs)
 
 
+def convert_coordinates(point):
+    if len(point) != 2:
+        raise KeyError('A point should only have two dimensions', point)
+    if 'lat' and 'lon' in point:
```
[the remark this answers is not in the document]
konvulse: You're not the first to mention that so now I'll have to think about it.  Before, I stated that I think the array is confusing and generally a bad way to represent something like x,y or lat,lon.  I think it makes it extra confusing, x,y --> 0,1; lat,lon --> 1,0.

api.py
```diff
@@ -367,7 +388,134 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+                coordinates.append(convert_coordinates(point))
+        if 'polygon' in sample:
+            if len(coordinates) > 0:
+                raise KeyError('A sample may only contain one of point, path or polygon', sample)
+            if len(sample['polygon']) < 3:
+                raise KeyError('polygons must have at least three points')
+            for point in sample['polygon']:
+                coordinates.append(convert_coordinates(point))
+
+        sample['_kind'] = sample['kind']
```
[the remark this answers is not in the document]
konvulse: It's a future proofing measure.  The idea is that we will want to eliminate the kind field in favor of an arbitrary attribute of the user's choosing.  So if we automatically add a `_kind` attribute we enable data ingested today to continue working.

api.py
```diff
@@ -367,7 +388,134 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+    requests.Response
+        Returns an error or the final response message when the job is no longer running.
+
+    Raises
+    ------
+    requests.HTTPError
+        Requests that result in an error raise an exception with information about the failure. See :py:func:`requests.Response.raise_for_status` for more information.
+    """
+
+    if not isinstance(sample_list, list):
```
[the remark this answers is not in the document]
konvulse: Hmm, that does seem weird.

I did it so that a user could ingest a single sample without adding it to a list.  Maybe it's not worth it.

api.py
```diff
@@ -388,6 +388,107 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
     return ingest_entities(dataset_id, {'entities': [entity]}, **kwargs)
 
 
+def convert_coordinates(point):
+    if len(point) != 2:
+        raise KeyError('A point should only have two dimensions', point)
```
clawconduce: `ValueError` is worth considering too

api.py
```diff
@@ -388,6 +388,107 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+    Raises
+    ------
+    requests.HTTPError
+        Requests that result in an error raise an exception with information about the failure. See :py:func:`requests.Response.raise_for_status` for more information.
+    """
+
+    if not isinstance(sample_list, list):
+        sample_list = [sample_list]
+
+    if 'id' and 'time' in sample_list[0]:
```
clawconduce: this seems delicate - only the first sample needs to have id and time.  I'd put this in `convert_samples_to_entity_set` and check every sample.

Also, I'd say why the sample was invalid - I'd want to know how to fix it w/o looking for this error msg in the code & seeing why it happened

api.py
```diff
@@ -427,6 +528,19 @@ def ingest_entities(dataset_id, data, **kwargs):
     return response
 
 
+def _ingest_entity_set(dataset_id, entity_set, **kwargs):
```
clawconduce: did you also want `ingest_entities` to call this?

util.py
```diff
@@ -208,6 +222,8 @@ def get_z_score(key, value):
     score = 0
     if key == 'z' or key == 'height' or key == 'depth' or key == 'altitude':
         score += 1000
+    if key == 'alt':
+        socre += 100
```
clawconduce: `score` instead of `socre`

api.py
```diff
@@ -388,6 +388,107 @@ def set_generic_data(dataset_id, key, data_string, **kwargs):
...
+def convert_samples_to_entity_set(sample_list):
+    conduce_keys = ['id', 'kind', 'time', 'point', 'path', 'polygon']
+    entities = []
+    for sample in sample_list:
+        if not isinstance(sample['time'], datetime):
+            raise TypeError('time must be a datetime object', sample['time'])
+        coordinates = []
+        if 'point' in sample:
+            coordinates = [convert_coordinates(sample['point'])]
+        if 'path' in sample:
```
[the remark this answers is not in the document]
konvulse: Understood.  I was hoping we could add support for that later, since it is an extension of the API and not a change.

util.py
konvulse: This is why this library needs unit tests.

util.py
konvulse: I did this to communicate that the function is private and should not be used outside the the module and to keep it from appearing in sphinx documentation.

Then I immediately broke that guideline by using it outside the module.  It seems like the right thing to do until we can get through all the callers of the old `ingest_entities`, now `_ingest_entity_set` and update them to use the correct public API.

Since we are not documenting the Conduce entity structure we can't include this function in our official API.

api.py
konvulse: Yes

api.py
konvulse: My intention here was not to validate every sample, but rather to validate that the user passed samples and not entities, my thinking being that the user would not confuse structures within a list.  They would either be samples or entities.

However maybe I should be validating every sample and providing detailed information about missing/invalid fields.

I could do something simple now, but that seems like a follow on story.

api.py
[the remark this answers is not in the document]
konvulse: Consider the use case where a user wants to ingest a single sample.  It seems arduous to make the user but the sample in a list.  The purpose is to provide convenience to the user.  If they have multiple samples, they obviously need to put them in a list.  If they have one sample don't make them put brackets around it, to them the input is not a list.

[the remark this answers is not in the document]
konvulse: I'll cave on this one.  If it's going to behave the way I coded it, the documentation should indicate as much.  It seems complicated to document that a function parameter can be one of many types.  However, I'd say that's the way python intends things to work.

If I were writing this in C++ I would override the function such that the user could pass a single struct or a list of structs.

api.py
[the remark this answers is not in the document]
chlela: I'm not convinced that this should part of this PR but ok, it's very low importance

api.py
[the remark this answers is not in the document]
chlela: I think it's fine to support two types in the same argument, compiled languages don't support it but it's fine in scripting languages.  Typescript supports the or operator so a function can accept multiple type or return multiple types, no issues.  

We should update the parameter name to be different than sample_list (since we support a single sample), update the header documentation for awarness and add a small line describing two supported types so when a developer is skimming through the code, it doesn't throw him off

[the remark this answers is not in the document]
konvulse: This conditions throws an error now.

api.py
konvulse: It's a perspective thing right, is it the value of the dict as a whole or the specific key that is in error.

I tried to be judicious about when I used KeyError or ValueError.

PR #45, opened 2017-11-01
doc create api key

konvulse:
This adds the documentation for creating api keys.  There is some cleanup work to be done in index.rst that is outside the scope of this story.

Please review this document:
https://conduce-conduce-python-api.readthedocs-hosted.com/en/feature-doc-create-api-key/api-key-creation.html 

and make comments here.

There are also code changes made to accommodate the documentation.

Exit criteria:

A non-Conduce expert can read this document and create an API key assuming they have a Conduce account and understand how to install a Python package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ If you provide an API key for your default user and host you will never have to 
 
 The CLI provides many high level functions.  These include:
 
+- Creating API keys
 - Adding and removing datasets
 - Creating a dataset from a CSV or JSON file
-- Adding generic data to a dataset entity
-- Listing datasets and orchestrations
+- Listing datasets and other resources
 
 All CLI functionality is available through `conduce/api.py` and all API functionality should be usable with or without a `.conduceconfig` file.
 
--- a/api.py
+++ b/api.py
@@ -693,7 +693,53 @@ def create_orchestration(name, orchestration_def, **kwargs):
     return create_json_resource('ORCHESTRATION', name, orchestration_def, **kwargs)
 
 
+def list_api_keys(**kwargs):
+    """
+    Get the users's API keys. 
+
+    Send an HTTP GET request to list this user's API keys.
+
+    Parameters
+    ----------
+    **kwargs
+        See :py:func:`make_get_request`
+
+    Returns
+    -------
+    requests.Response
+        Returns a list of API key dictionaries.
+
+    Raises
+    ------
+    requests.HTTPError
+        Requests that result in an error raise an exception with information about the failure. See :py:func:`requests.Response.raise_for_status` for more information.
+    """
+
+    response = make_get_request('apikeys/list', **kwargs)
+    return json.loads(response.content)
+
+
 def create_api_key(**kwargs):
+    """
+    Add an API key to a user's account.
+
+    Send an HTTP POST request for a new API key.  The API key is added to the account of the user who made the request.
+
+    Parameters
+    ----------
+    **kwargs
+        See :py:func:`make_post_request`
+
+    Returns
+    -------
+    requests.Response
+        Returns the string representation of the new API key.
+
+    Raises
+    ------
+    requests.HTTPError
+        Requests that result in an error raise an exception with information about the failure. See :py:func:`requests.Response.raise_for_status` for more information.
+    """
     response = make_post_request(
         {"description": "Generated and used by conduce-python-api"}, 'apikeys/create', **kwargs)
     return json.loads(response.content)['apikey']
--- a/api.py
+++ b/api.py
@@ -695,7 +695,7 @@ def create_orchestration(name, orchestration_def, **kwargs):
 
 def list_api_keys(**kwargs):
     """
-    Get the users's API keys. 
+    Get this users's API keys.
 
     Send an HTTP GET request to list this user's API keys.
 
@@ -721,7 +721,7 @@ def list_api_keys(**kwargs):
 
 def create_api_key(**kwargs):
     """
-    Add an API key to a user's account.
+    Add an API key to this user's account.
 
     Send an HTTP POST request for a new API key.  The API key is added to the account of the user who made the request.
 
--- a/doc/source/api-key-creation.rst
+++ b/doc/source/api-key-creation.rst
@@ -4,7 +4,7 @@
 API Key Creation
 ====================
 
-API keys are used to authenticate requests made using the Conduce REST API.  API keys are created by the user and should be protected like passwords.  Anyone with an API key can make requests to Conduce and perform operations as the user who requested the key.
+API keys are used to authenticate requests made using the Conduce REST API.  API keys are created by the user and should be protected like passwords.  Anyone with an API key can make requests to Conduce and perform operations as the user who owns the key.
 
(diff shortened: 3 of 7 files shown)


doc/source/api-key-creation.rst
```diff
@@ -0,0 +1,37 @@
+.. _api-key-creation:
+
+====================
+API Key Creation
+====================
+
+API keys are used to authenticate requests made using the Conduce REST API.  API keys are created by the user and should be protected like passwords.  Anyone with an API key can make requests to Conduce and perform operations as the user who requested the key.
```
jrmConduce: s/who requested the key/who created the key

doc/source/api-key-creation.rst
```diff
@@ -0,0 +1,37 @@
...
+**username/password**
+    used to establish a session
+**session cookie**
+    returned from a user name password login.  Used to continue a session started with a username/password login.
+**API key**
+    a unique string generated upon request that can be used to authenticate requests.
+
+In order to generate an API key, you must have a Conduce account.  Contact support@conduce.com if you do not already have an account.  Once your account is registered, you may create an API key.  It is easiest to do this using the Conduce command line utility.  Once you have :ref:`installed the Conduce Python API <https://github.com/ConduceInc/conduce-python-api>` execute the following command from your shell prompt::
+
+    conduce-api create-api-key --user=new_user@conduce.com --host=app.conduce.com
```
jrmConduce: You should explicitly tell the reader to replace this email address with the email address they used to create their account

jrmConduce: It would be nice if we could find a way to prevent the user from specifying the host.  A user will only interact with a single host, so this is just wasteful for them to type.

doc/source/api-key-creation.rst
```diff
@@ -0,0 +1,37 @@
...
+
+After you submit your password, the new API key will be printed to the prompt and the command will finish.  Alternatively you may run a Python REPL and call :py:func:`api.create_api_key` directly.::
```
jrmConduce: You should tell the reader the they will be prompted for a password.  You mention it below with "again", but you only talk about submitting here.

doc/source/api-key-creation.rst
```diff
@@ -0,0 +1,37 @@
...
+Again, you will be prompted for your password.  Upon submission, the new API key will be written to the prompt.
+
+.. CAUTION::
+    API keys can be used to perform any user operation.  This includes access to private data and personal information.  Be sure to protect API keys accordingly.
+
+API keys you have created may be retreived using the following command::
+
+    conduce-api list-api-keys --user=new_user@conduce.com host=app.conduce.com
+
+The API keys and the date on which they were created will be printed to the terminal.  As with API creation.  The API keys may be listed using the Python API directly using :py:func:`api.list_api_keys`.
```
jrmConduce: You should also include an example of using their new API key.  Otherwise what was the point?

doc/source/api-key-creation.rst
konvulse: Am I being too literal.  I am distinguishing the user who requests that a key be created from the system that actually creates it.  I don't want the user to get the impression that they must create an API key and POST it to Conduce.

konvulse: But I think I see your point.  What about "the user who owns the key?"

doc/source/api-key-creation.rst
konvulse: They do not have to specify the host if they have a `~/.conduceconfig` or an email address for that matter.

I cover that in "Getting Started" (README.md in GitHub).

konvulse: Is there a string I could use in place of `--user=new_user@conduce.com` that indicates that the user could enter their email address?  Maybe `--user=<email-address>`

doc/source/api-key-creation.rst
konvulse: I was trying not to be overly wordy.  I was hoping that the prompt would be implied by submission.

I guess since I mention prompting below, I should move that language here and skip it there.

doc/source/api-key-creation.rst
konvulse: Well that just makes sense 😖

doc/source/api-key-creation.rst
jrmConduce: sure

doc/source/api-key-creation.rst
jrmConduce: If it's already documented elsewhere, I suggest you make a statement indicating something like "the following examples assume you have already configured the system as specified in the getting started guide" (with a link)

And yes to the `--user=<email-address>`

doc/source/api-key-creation.rst
jrmConduce: yeah... you can't say "again" below if you didn't mention it here first.  And this is the first place they will encounter this, so you should be more descriptive on the initial one

doc/source/getting-started.rst
```diff
@@ -0,0 +1,7 @@
+.. _getting-started:
+
+===============
+Getting Started
+===============
+
+.. mdinclude:: ../../README.md
```
clawconduce: I gave the README a quick skim - it mentions a list of features this project provides & includes a list of features that mentions listing orchestrations.  It might be easier to say "list users resources, such as datasets" and avoid using the word orchestration

api.py
```diff
@@ -693,7 +693,53 @@ def create_orchestration(name, orchestration_def, **kwargs):
...
+        Requests that result in an error raise an exception with information about the failure. See :py:func:`requests.Response.raise_for_status` for more information.
+    """
+
+    response = make_get_request('apikeys/list', **kwargs)
+    return json.loads(response.content)
+
+
 def create_api_key(**kwargs):
+    """
+    Add an API key to a user's account.
```
clawconduce: supernit: I'd say "Add an API key to *this* user's account."

requirements.txt
```diff
@@ -1,5 +1,6 @@
 retrying
 python-dateutil
 Sphinx
+m2r
```
clawconduce: this will need a rebase, this should go into `requirements-docs.txt`

doc/source/getting-started.rst
konvulse: Good point.  I should have looked over the readme more carefully.

api.py
konvulse: correct -- it's documentation, that stuff is important

README.md
```diff
@@ -63,10 +63,10 @@ If you provide an API key for your default user and host you will never have to
 
 The CLI provides many high level functions.  These include:
 
+- Creating API keys
 - Adding and removing datasets
 - Creating a dataset from a CSV or JSON file
-- Adding generic data to a dataset entity
```
konvulse: We shouldn't advertise generic data stuff.

README.md
```diff
@@ -63,10 +63,10 @@ If you provide an API key for your default user and host you will never have to
...
-- Listing datasets and orchestrations
```
konvulse: We don't need to say orchestrations here.

PR #46, opened 2017-11-01
split requirements

konvulse:
Splitting runtime requirements from documentation requirements.  ReadTheDocs should be updated to use `requirements-all.txt` after this is merged.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,0 +1,2 @@
+Sphinx
+m2r
--- a/README.md
+++ b/README.md
@@ -86,3 +86,34 @@ email@example.com:
 ```
 
 and is stored at `~/.conduceconfig`
+
+## Building documentation
+
+Documentation is built with Sphinx.  To install Sphinx and Sphinx extensions run:
+
+```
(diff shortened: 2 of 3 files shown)


requirements-docs.txt
```diff
@@ -0,0 +1,2 @@
+Sphinx
```
clawconduce: after trying to test this I realized that this doesn't actually remove anything from `requirements.txt` - I think this should also remove `Sphinx` from `requirements.txt`

clawconduce:
btw I tested this using `./venv/bin/pip install git+https://github.com/ConduceInc/conduce-python-api@eaf2f188975c256a4457f9933dd75a9fdf384128` the hash will be different after changes.

Branch name might work instead of commit, but the output wasn't clear so I used commit hash which definitely was used

requirements-docs.txt
konvulse: I can't do that yet, because it will break readthedocs.

Once this is merged I'll update readthedocs to point to the new requirements file and remove the documentation requirements from requirements.txt.

PR #48, opened 2017-11-02
Removed sphinx from requirements.txt

konvulse:


clawconduce:
btw, `./venv/bin/pip install git+https://github.com/ConduceInc/conduce-python-api@charlie/remove-doc-requirements` does work - not much installs with that command 👍

PR #52, opened 2017-11-06
Feature/working samples

clawconduce:


util.py
```diff
@@ -238,9 +238,13 @@ def get_default(field):
...
     elif field == 'timestamp_ms':
-        return -2e48 - 1
+        #this int64 field will be a float64 sometimes too (such as in the VBO
+        #that is rendered in the frontend) and we want to keep the sub-second
+        #precision. There are 52 bits in the mantissa, so under that by some margin
+        return -2**48 - 1
     elif field == 'endtime_ms':
-        return 2e48 - 1
+        #See timestamp_ms notes about float64
+        return 2**48
```
konvulse: The idea was that these would be the same value such that when defaulted `timestamp * -1 == endtime_ms`.  This should be 2e48-1.

util.py
```diff
@@ -238,9 +238,13 @@ def get_default(field):
     if field == 'identity':
         return uuid.uuid4()
     elif field == 'timestamp_ms':
-        return -2e48 - 1
+        #this int64 field will be a float64 sometimes too (such as in the VBO
+        #that is rendered in the frontend) and we want to keep the sub-second
+        #precision. There are 52 bits in the mantissa, so under that by some margin
+        return -2**48 - 1
```
konvulse: Man I really screwed up! This should be `-(2**48-1)`.

util.py
```diff
@@ -238,9 +238,13 @@ def get_default(field):
     if field == 'identity':
         return uuid.uuid4()
     elif field == 'timestamp_ms':
-        return -2e48 - 1
+        #this int64 field will be a float64 sometimes too (such as in the VBO
```
konvulse: I don't think this comment accurately explains why we chose this number.  It's just a big number that will have the same value whether represented as a floating point number or a signed int.  If we were going to document something, it might be "a magic number that represents positive/negative infinity."

Having said that, I don't think this should be documented here.  This is a client helper library that is just abiding by the rules of the API.  There will be API documentation that discusses timeless entities and how they are represented.  So rather than saying, "because the API works this way," I think we can just leave the comment out.

elliottconduce:
Woop! This fixes my remaining issues with building DHL for testing.

setup.py
```diff
@@ -8,7 +8,7 @@
 
 setup(
     name="conduce",
-    version="2.0.3",
+    version="2.0.4",
```
konvulse: Ah, excellent!

PR #53, opened 2017-11-07
SKW-2013 add remove_api_key method

rclancey:


api.py
```diff
@@ -870,6 +870,35 @@ def create_api_key(**kwargs):
     return json.loads(response.content)['apikey']
 
 
+def delete_api_key(key, **kwargs):
```
konvulse: Method should be `remove_api_key` to be consistent with the rest of the API.

api.py
```diff
@@ -870,6 +870,35 @@ def create_api_key(**kwargs):
...
+    """
+    Delete an API key from this user's account.
+
+    Send an HTTP POST request to delete the given API key.  The key in question may not be used for authentication of the request.
```
konvulse: Is the key destroyed, or does Conduce hang on to it?

If it destroyed, we should mention that here.  Otherwise, lets update the language to say:

>The key in question may no longer be used for authentication of the request.

api.py
```diff
@@ -870,6 +870,35 @@ def create_api_key(**kwargs):
...
+    Returns
+    -------
+    Nothing
+
+    Raises
+    ------
+    requests.HTTPError
+        Requests that result in an error raise an exception with information about the failure. See :py:func:`requests.Response.raise_for_status` for more information.
+    """
+    if key == kwargs.get('api_key'):
```
konvulse: I suggest we remove this logic and allow users to use an API key to delete/remove the same API key.

api.py
rclancey: This is one of those methods that actually deletes something.  So when you call this method, the key is lost forever and cannot be recovered or restored.

api.py
rclancey: I put in this restriction (it doesn't exist in the REST API) to avoid accidents.  I'm happy to take it out if you're comfortable with the possibility of users accidentally deleting the wrong key.  Would it be acceptable to remove this check from this function but keep it in the command-line version?

PR #55, opened 2017-11-07
SKW-1982 updated ingest_entities to use new entity list rather than e…

rclancey:
…ntity set

### DIFF
--- a/api.py
+++ b/api.py
@@ -509,6 +509,7 @@ def ingest_samples(dataset_id, sample_list, **kwargs):
 def convert_entities_to_entity_set(entity_list):
     conduce_keys = ['id', 'kind', 'point', 'path', 'polygon']
     entities = []
+    ids = set()
     for idx, ent in enumerate(entity_list):
         if not 'id' in ent:
             raise ValueError('Error processing entity at index {}. Entities must include an ID.'.format(idx), ent)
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -533,10 +567,9 @@ def ingest_entities(dataset_id, data, **kwargs):
         Requests that result in an error raise an exception with information about the failure. See :py:func:`requests.Response.raise_for_status` for more information.
     """
 
-    if isinstance(data, list):
-        data = {'entities': data}
+    entity_set = convert_entities_to_entity_set(entity_list)
```
konvulse: Wondering if we should do some checks here.  For instance, require no more than one entity per ID, since they will just overwrite one another.  Are there others?

api.py
```diff
@@ -506,7 +506,41 @@ def ingest_samples(dataset_id, sample_list, **kwargs):
...
+            'timestamp_ms': util.get_default('timestamp_ms'),
+            'endtime_ms': util.get_default('endtime_ms'),
+            'path': coordinates,
+            'attrs': attributes
+        })
+
+    return {'entities': entities}
+
+
+def ingest_entities(dataset_id, entity_list, **kwargs):
```
konvulse: I'll update the documentation here.

konvulse:
Should we also do a check to ensure that the user did not populate a `time` field?

api.py
```diff
@@ -506,18 +506,73 @@ def ingest_samples(dataset_id, sample_list, **kwargs):
...
     Upload :ref:`entities <conduce-entities>` to the Conduce datastore.
 
     A convenience method that adds entities to the Conduce datastore and waits for the job to complete. This function POSTs an entity set to Conduce and :py:func:`wait_for_job` until the ingest job completes.
 
+    Each element of ``entity_list`` must have the following structure::
+
+        {
+            "id": <string>,
+            "kind": <string>,
+            "point": {
```
jrmConduce: Are we only allowing points at this time?

konvulse: No, we also allow paths and polygons.  However, I don't want to document those alternate structures here.  Would you be satisfied if I changed the language from "must" to something that indicates this is an example of a valid structure.

I link to the conduce-entities document above so I don't need to do it again here, I think.

konvulse: I've made changes to address your comment.

PR #57, opened 2017-11-09
Basic data ingest

konvulse:
This change is intended to provide more fluid documentation.  Starting from the main page the user to guided to documents that help with installation of the API and setting up an API key.  Then the user is presented with documentation on ingesting data.


api.py
```diff
@@ -701,22 +812,57 @@ def create_substrate(name, substrate_def, **kwargs):
...
+    See :py:func:`find_resource` for more information on the other query parameters.
+    """
     return find_resource(type='LENS_TEMPLATE', **kwargs)
 
 
 def find_dataset(**kwargs):
+    """
+    Find datasets matching the given parameters.
+
+    A convenience methond that calls :py:func:`find_resource` with the ``type`` argument set to "DATASET."
```
chlela: method instead of methond

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+====================
+Intro to Data Ingest
+====================
+
+Visualizing your data in Conduce begins with ingest.  Ingest is the process translating your data into a format that Conduce understands.  Conduce uses entities to define unique objects.  Entities are ingested into datasets.  Datasets hold collections of related entities and provide the interface a user needs to connect their data to a visualization renderer.
+
+Conduce needs a few key pieces of information in order to visualize your data in space and time, those are:
+
+**Identity**
+   Identities distinguish one entity from another.  An entity exists in Conduce as one more more samples.  All samples that describe the same entity share an identity.
```
chlela: extra more in `more more`

README.md
```diff
@@ -4,34 +4,57 @@ This library implements the most commonly used Conduce API endpoints.  It can be
...
+# Using the API
+
+Once you have installed the API, you may use it in a python script by importing the `conduce` module:
+```
+import conduce
+```
+Then you may use API functions, like `create_dataset` as follows:
 ```
-python setup.py install
+conduce.api.create_dataset('my-dataset-name', host='app.condcuce.com', user='email@example.com')
```
jrmConduce: Stupid question... if the api lives inside an api package, what, if anything, lives directly in the main conduce package?  It seems like an unnecessary indirection when looking through the lens of "this is the conduce api library".

README.md
```diff
@@ -4,34 +4,57 @@ This library implements the most commonly used Conduce API endpoints.  It can be
...
 ```
+However, this requires that the user authenticate with their password when the script is executed.  To avoid this, provide an `api-key` argument instead of `user`.  If you would like to learn more about API key creation, take a look at [API Key Creation](https://conduce-conduce-python-api.readthedocs-hosted.com/en/latest/api-key-creation.html).
+
+The most common use for the python API is data ingest.  Data ingest is too large a topic to be covered here, if you'd like to learn more check out [Intro to Data Ingest](https://conduce-conduce-python-api.readthedocs-hosted.com/en/latest/data-ingest.html)
```
jrmConduce: The wording here feels too casual.  "take a look at", "check out", etc.

api.py
```diff
@@ -45,22 +45,66 @@ def new_func(*args, **kwargs):
...
+    A convenience method that calls :py:func:`list_resources` with the ``type`` field set to "DATASET."
+
+    See :py:func:`find_resource` for information on return type.
+    """
     return list_resources('DATASET', **kwargs)
 
 
 def list_substrates(**kwargs):
+    """
+    List all accessible substrates.
```
jrmConduce: Do we have a definitions/concepts section yet?  We certainly need to explain substrates somewhere

doc/source/conduce-entities.rst
```diff
@@ -9,9 +9,22 @@
...
 ================
 
-.. note:: WIP
 
-Conduce uses a concept called entity to define objects that exist in space and time.  An entity can exist for an instant or across a time span.  It can describe a point in space, a path (sequence of points), or a region (polygon).  An entity is defined by a series of samples that describe it's location in space over a period of time.  An entity also has attributes that describe it's characteristics. An entity's attributes can change over time.
+------------------
+What is an entity?
+------------------
+
+Conduce uses a concept called entity to define a person, place, or thing that exist in space and time.  An entity can exist for an instant or across a time span.  It can describe a point in space, a path (sequence of points), or a region (polygon).  An entity is defined by a series of samples that describe it's location in space over a period of time.  An entity also has attributes that describe it's characteristics. An entity's attributes can change over time.  Entities can be static or dynamic.  To understand the difference between the two, it is helpful to understand dynamic entities first.
```
jrmConduce: s/exist/exists

doc/source/conduce-entities.rst
```diff
@@ -9,9 +9,22 @@
...
+
+++++++++++++++++
+Dynamic entities
+++++++++++++++++
+Dynamic entities are composed of a sequence of samples.  The term sample is taken from the `signal processing domain <https://en.wikipedia.org/wiki/Sampling_(signal_processing)>`_ and describes the state of an entity at a discrete moment in time.  A sample describes an entity's location in space and other attributes.  These qualities can change from sample to sample so that when a visualization is constructed, these state changes can be visualized.  A sequence of samples that share the same ID define the entity, and so, an entity is a sequence of samples.  Entities defined this way are said to be `dynamic`.  Conduce does not require that an entity's sampling interval be regular.  This allows users to update entities based on events (state changes) rather than period (time changes).  However, it should be considered that some types of entities should be sampled regularly in order generate an accurate visual representations.
```
jrmConduce: As long as we're referencing signal processing, should we replace `dynamic` with `discrete`?

jrmConduce: The phrasing "it should be considered" sounds awkward to me.  Perhaps "it should be noted" instead?  Or perhaps make it active instead of passive: "However, note that some types..."

doc/source/conduce-entities.rst
```diff
@@ -9,9 +9,22 @@
...
+
++++++++++++++++
+Static entities
++++++++++++++++
+A static entity is one that does not change over time.  If an entity has a state that does not change, it may be ingested with a single message that does not have a time field.  As such the entity is a sequence with one element that defines the entity for all time.  Entities defined by a single state are considered `static`.  This document focuses on dynamic entities.  However the same basic principles apply for ingest static entities, with the exception that static entities do not have ``time`` fields and, because the entity exists across all time, only one instance of the entity can exist in the dataset.  See :doc:`data-ingest` for an example of ingesting static entities.
```
jrmConduce: My biggest objection to using `discrete` for samples (which I think is a better term than dynamic) is that static entities aren't really continuous.

jrmConduce: s/However the same basic principles apply for ingest static entities/However the same basic principles apply for ingesting static entites

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+.. toctree::
+    :hidden:
+
+    conduce-entities
+
+====================
+Intro to Data Ingest
+====================
+
+Visualizing your data in Conduce begins with ingest.  Ingest is the process translating your data into a format that Conduce understands.  Conduce uses entities to define unique objects.  Entities are ingested into datasets.  Datasets hold collections of related entities and provide the interface a user needs to connect their data to a visualization renderer.
```
jrmConduce: "Ingest is the process translating your data" isn't quite right.
"process of", or "process that translates" would be better

jrmConduce: If you're going to use the term "visualization renderer" here, you need to explain the rendering concept somewhere.  Probably in a yet-to-be-written section defining terms and giving an overview of the conduce system

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Conduce needs a few key pieces of information in order to visualize your data in space and time, those are:
+
+**Identity**
+   Identities distinguish one entity from another.  An entity exists in Conduce as one more more samples.  All samples that describe the same entity share an identity.
+**Location**
+   An entities location describes where in space the entity exists.  Locations are defined in units according to the coordinate system in which they are displayed. (Latitude and Longitude on a world map.)
```
jrmConduce: s/entities/entity's

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+**Time**
+   Time defines the moment in time at which the rest of the information describing an entity is valid.
```
jrmConduce: Several comments on time

1. You should remind the reader that time is only required for samples and is implicitly infinite for entities
2. You should point out that a sample is uniquely identified by its identity and time.  That is, only one sample may be specified with a given identity and time
3. Since time is more fundamental than location, it might be worth putting it second after identity

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+
+With this information Conduce can build a representation of an object that exists in space and time.  Entity definitions can be more complex, for more information on how to build entities please read :doc:`conduce-entities`.
+
+--------
+Datasets
+--------
+
+Datasets are the containers that hold entities.  A dataset may be thought of as a table.  Each entity in a dataset is similar in nature, but has unique properties.  In order to represent and entity in Conduce, its sample data must be ingested into a dataset.  In order to perform any API operations it is recommended that you generate an API key.  (See :doc:`api-key-creation` for more information.)  The first step in ingesting data is acquiring a dataset ID.  If you are ingesting data into a new dataset you'll need to create the dataset  before entities can be ingested.
```
jrmConduce: s/In order to represent **and** entity/In order to represent **an** entity

jrmConduce: "its sample data must be ingested into a dataset"

Should this reference samples explicitly?

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+
+++++++
+Ingest
+++++++
+
+Datasets are populated through a process called ingest.  If creating or updating static entities call :py:func:`api.ingest_entities`.  If you are creating or updating dynamic entities call :py:func:`api.ingest_samples`.
+
+For our example, you are ingesting entities and will use :py:func:`api.ingest_entities`.  The list of entities you generated was written to a variable named ``entity_list``.  All that's left is to call the API function using the dataset ID you created earlier and the API key you generated::
+
+    conduce.api.ingest_samples(world_cities_dataset_id, entity_list, host=app.conduce.com, api-key=00000000-0000-0000-0000-000000000000)
```
jrmConduce: Shouldn't this be ingest_entities?

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+   Identities distinguish one entity from another.  An entity exists in Conduce as one more more samples.  All samples that describe the same entity share an identity.
+**Location**
+   An entities location describes where in space the entity exists.  Locations are defined in units according to the coordinate system in which they are displayed. (Latitude and Longitude on a world map.)
+**Time**
+   Time defines the moment in time at which the rest of the information describing an entity is valid.
+
+With this information Conduce can build a representation of an object that exists in space and time.  Entity definitions can be more complex, for more information on how to build entities please read :doc:`conduce-entities`.
+
+--------
+Datasets
```
jrmConduce: I would change this header to "Creating a Dataset"

The section is sort of describing what a dataset is, but it's really giving the reader instructions on how to create it.  The following sections are focused on the the steps the user takes.  This one should too

doc/source/data-ingest.rst
clawconduce: and the first "more" should be an "or"

doc/source/data-ingest.rst
clawconduce: originally +1 to all, esp. 2 & 3

Though I should also point out, time is only included for sample data, not for static data

doc/source/data-ingest.rst
clawconduce: "dataset  before" - 2 spaces - does 2 spaces matter?

clawconduce: The API key sentence should go at the end - it threw me off while going thru the paragraph.  I would add this to the end of the paragraph:

The example below ingests data via the API.  In order to perform API operations ...

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+       "version": 0,
+       "mime": "application/json",
+       "modify_time": 1508869186664,
+       "revision": 4,
+       "type": "DATASET",
+       "id": "a262342b-22d2-4368-4698-cf75b62b7cb3",
+       "size": 2
+     }
+
+This dataset object contains information to help identify a dataset.  The fields ``name`` and ``tags``, along with ``create_time`` and ``modify_time`` are particularly useful for identifying datasets.  The :py:func:`api.find_datasets` function can be used to list only datasets that match certain search criteria.
```
clawconduce: it might be worth reiterating that only the uuid is truly unique - there can be 2 datasets with the same name, tag, & less likely so, the same create_time & modify_time.

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+
+-------------------
+Populating datasets
+-------------------
+
+Once a dataset has been created, it needs to be populated before it can be visualized.  Datasets are populated by making API calls to create entities.  An entity is a person, place, or thing that exists in space and time.  See :doc:`conduce-entities` for detailed information on creating entities.
```
clawconduce: I think this is confusing.  We don't create entities, we add samples.  This makes me think as a user I'll have to run a create entities api call or something.

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+- lng
+- pop
+- country
+- iso2
+- iso3
+- province
+
+These fields are described in detail on the page from which the data is downloaded.  For this example you'll work with all the fields.  However the only fields that are required are ``lat`` and ``lng``.
+
+An entity requires the following fields in order to be ingested: ID, location, and kind.  Kind is a field that describes the type of thing the entity represents.  you'll be ingesting this data as static entities, so a timestamp is not required.  you'll map the source data to Conduce entities as follows::
```
clawconduce: you'll --> You'll

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+        "name": <city>,
+        "long_name": <city_ascii>,
+        "population": <pop>,
+        "province": <province>,
+        "country": <country>,
+        "iso2": <iso2>,
+        "iso3": <iso3>,
+    }
+
+Notice that two of the fields ``id`` and ``kind`` are not derived from the data.  you will generate a unique ID just in case any of the city names are the same.  you will hardcode ``kind`` to city in case you want to ingest other types of data into this dataset in the future.  If you wanted to do something more sophisticated you could categorize the cities by population and set ``kind`` to something like "small_city," "medium_city," and "large_city."
```
clawconduce: you --> You (in multiple places)

doc/source/data-ingest.rst
```diff
@@ -0,0 +1,140 @@
...
+        "population": 2997,
+        "province": "Badghis",
+        "country": "Afghanistan",
+        "iso2": "AF",
+        "iso3": "AFG",
+    }
+
+Writing the code to iterate over the CSV file and convert each record to a Conduce entity is left to the reader.  However there are utilities included with the Conduce Python API that provide example implementations.
+
+The resulting entities should be compiled into a list.  Once you have generated the entity list it is time to ingest the data to Conduce.
```
clawconduce: "ingest the data to Conduce" sounds awkward to me - maybe "into Conduce"?  I can't put my finger on it though

doc/source/data-ingest.rst
clawconduce: Maybe this should say "Static datasets are populated by making API calls to add entities.  For dynamic datasets, there is no need to create entities - instead users make API calls to add samples"

doc/source/data-ingest.rst
clawconduce: Giving this a 2nd pass - is this too static dataset focused?  If I were to explain this to DHL talking about CMX data, I don't think this would make sense.  Can we remove the reference to entities.

Also, we reference samples below, but at this point in the docs, the user probably has not seen the entities page - we should give a quick definition for sample and entity

doc/source/data-ingest.rst
clawconduce: The user doesn't know what a sample is at this point

doc/source/data-ingest.rst
clawconduce: I'd reword "entity is valid" --> "entity was captured"

clawconduce:
FYI, I saw several of my comments get attached to other reviews - some of those were good, like a reference to samples before the user knows what a sample is

README.md
konvulse: Are you talking about `__init__.py`?

`conduce.py` is the command line interface.

README.md
konvulse: I was going for casual.

api.py
konvulse: Eventually.  It's not part of data ingest.

doc/source/data-ingest.rst
konvulse: RE sample definition, I'll think about that.

doc/source/conduce-entities.rst
konvulse: any time you can replace four words with one, that's a win

doc/source/data-ingest.rst
konvulse: I agree with your points. I think that I make them at various points (redundantly even) throughout the documentation.  For definitions, the key is to find wording that is brief and understandable yet meaningful and complete.

I debated whether to include this definitions section in this document at all.  Now I'm lean toward no.

konvulse: I think "captured" assumes too much.  I don't know when (or if) the data was captured and that term does not communicate how conduce interprets the data.

doc/source/data-ingest.rst
konvulse: Not sure about the overview comment.  I'm not going for completeness.  I do think it is important to hint at what comes next where appropriate.  However this is a data ingest intro, not an everything conduce can do manual.  I think we'll have to let "visualization renderers" be a mystery a little longer.

konvulse: I agree that this is choppy claw.  I'll see about rewording.  This document focuses on static entities because they are the simplest to ingest.  The conduce entities goes through an example of ingesting dynamic entities.

doc/source/data-ingest.rst
konvulse: I'll make it an admonition.

konvulse: Regarding this phrase, "its sample data must be ingested into a dataset," I agonized over that for quite a bit.

If an entity is not defined by a sample, then what is the thing that defines it?  Entity is not a good answer.

README.md
jrmConduce: Just wondering why it's `conduce.api.create_dataset` and not `conduce.create_dataset`

doc/source/data-ingest.rst
konvulse: Weird, thought I fixed that already.

doc/source/data-ingest.rst
jrmConduce: Well... in our limited API, an entity is defined over all time via an ID, a location, and a set of attributes, or it is defined as a series of samples, each of which contains an ID, a location, a time, and a set of attributes.

The point is that there are 2 distinct ways to define an entity, so the documentation shouldn't give users the impression that there is only one in some places and 2 in other places.

README.md
```diff
@@ -4,34 +4,57 @@ This library implements the most commonly used Conduce API endpoints.  It can be
 
 `api.py` is the majority of the implementation.  `conduce.py` is a command line utility for invoking the API.  It also servers as a representative implementation of the library.
 
+# Prerequisites
+
+Conduce Python API requires that you have installed the following:
+
+- Python 2.7: https://www.python.org/downloads/
```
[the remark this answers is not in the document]
konvulse: Yes, that has been on my mind.

api.py
```diff
@@ -163,9 +207,9 @@ def make_delete_request(fragment, **kwargs):
         host : string
             The Conduce server's hostname (ex. app.conduce.com)
         api_key : string
-            The user's API key (UUID).  The user should provide `api_key` or `user` but not both.  If the user provides both `api_key` takes precident.
+            The user's API key (UUID).  The user should provide `api_key` or `user` but not both.  If the user provides both `api_key` takes precedent.
         user : string
-            The user's email address.  Used to look up an API key from the Conduce config or, if not found, authenticate via password.  Ignored if `api_key` is provided.
+            The user's email address.  Used to look up an API key from the Conduce configuration or, if not found, authenticate via password.  Ignored if `api_key` is provided.
```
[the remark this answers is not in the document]
konvulse: I was using the git convention.  This is not linux specific.

doc/source/data-ingest.rst
konvulse: What if I change create to add.  I agree we are not really creating.  Entities are created by the user and added to the dataset.

doc/source/data-ingest.rst
konvulse: Agreed.  I have a hard time with this language.  Send data to Conduce is certainly good.  I suppose the problem is that I am using "Conduce" to refer to the datastore.  I could say "into Conduce's datastore."

doc/source/data-ingest.rst
konvulse: I think there is a terminology problem.  Entities are the concept.  They are not directly represented in the dataset.  Yet I am using language that implies that is the case.  In reality only samples exist in the dataset.  Samples define the entity.  However we've restricted use of the word sample to only those objects that have a time component.  That leaves us with the question of what to call the object that defines a timeless/infinite entity.

Any ideas?

doc/source/data-ingest.rst
konvulse: ingest into sounds weird anyway.  I ingest something.  Someone does not ingest something into me.

So I suppose the correct phrasing would actually be, "send the data to Conduce to be ingested."   Very wordy.

konvulse: That probably means the `ingest_*` API are not named correctly 😿

konvulse:
Addressed lots of comments.

doc/source/data-ingest.rst
clawconduce: Yea, that may be the case.

Consume is very similar to ingest, and I use that word a lot more.  I think the issue is that Conduce does the consuming, and if sentence has the user or the data doing the ingesting, it sounds off.

Here the user is doing it.  I think the last example sounds good, but agree it is wordy.  It could be, "send the data to Conduce to ingest".

doc/source/data-ingest.rst
clawconduce: I can't think of any great words, though maybe something like static/timeless/infinite ___ works - static entity, static sample, infinite sample?

Alternatively, we could rename entity.  I recall something along the lines of "we represent unique objects using entities" somewhere, we could just call them objects.  Objects may be ok for the timeless things too, but I don't think it captures the timeless part well.

doc/source/data-ingest.rst
clawconduce: re-tagging this - not sure if it was caught & is now hidden.

doc/source/conduce-entities.rst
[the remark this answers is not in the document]
clawconduce: re-tagging this - not sure if it was caught & is now hidden.

api.py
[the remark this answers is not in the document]
clawconduce: ah yes, I forgot git uses that convention

doc/source/data-ingest.rst
clawconduce: The intro says "a constant continuous function (constant entity)" - something related to that maybe?

doc/source/data-ingest.rst
konvulse: I considered it.  I want to keep the focus on ingest.  Details of resources and their metadata seem off topic.  I did want to provide some information to help a user employ the commands the document recommends.

PR #58, opened 2017-11-13
Make changes easy data ingest from csv

konvulse:
Debug mode: one entity at a time
Specify specific values for kind and whether or not to generate IDs
Fix bug that led to "Nan" being interpreted as a double
Use unicode string encoding


konvulse:
This command should create a new populated dataset:

```
python conduce.py create-dataset world-cities --csv samples/simplemaps-worldcities-basic.csv --generate-ids --kind=city --host=dev-app.conduce.com
```

konvulse:
Results can be viewed here:

https://dev-app.conduce.com/#/orch/74fff48f-c6d4-4f9e-84a4-d195ad98a506

util.py
```diff
@@ -283,12 +290,15 @@ def get_attributes(attribute_keys, raw_entity):
...
+def score_fields(raw_entities, keys, **kwargs):
     # TODO: Use up to 100 values to score fields
     key_scores = {}
     for key in keys:
         key_scores[key] = {}
-        key_scores[key]['identity_score'] = get_id_score(key, raw_entities[0][key])
+        if kwargs.get('generate_ids', False):
+            key_scores[key]['identity_score'] = 0
+        else:
+            key_scores[key]['identity_score'] = get_id_score(key, raw_entities[0][key])
```
clawconduce: do you also want to set kind_score to 0 if kwargs.get("kind") is set?

konvulse: Admittedly, my solution for setting kind explicitly was a little hacky.

It's not necessary because unlike 'generate_ids' which forces default values, 'kind' sets an explicit value.  There is no need to manipulate the score in this case (a score of zero causes the default value to be used) we simply override whatever field the scoring algorithm chooses.  In this particular case I think 'kind' was scored to zero, but the logic in `get_field_value` is explicit.  The first step is use `override_value` if it exists.

PR #59, opened 2017-11-13
Add the version number to the help message

jrmConduce:


conduce.py
```diff
@@ -276,8 +276,10 @@ def get_entity(args):
 
 def main():
     import argparse
+    import pkg_resources
+    version = pkg_resources.require("conduce")[0].version
```
clawconduce: nitish - it'd be nice if this used a fallback version like "dev" or "source" if the package wasn't installed.

Essentially I'd hate for this util to break if getting the version number was the only thing not working

jrmConduce: I'm not sure if it will break.
I just ran this sequence with no issues:
```
mkdir venv
virtualenv venv
bin/venv/activate
pip install -r requirements.txt
python conduce.py -h
```
It printed the version number out of setup.py with no issues.

clawconduce: cool, I just ran it on my machine and saw similar behaviour (I saw 0.1.0 which is probably a dummy value)

PR #61, opened 2017-11-14
Feature/working generic data

clawconduce:


api.py
```diff
@@ -112,7 +112,11 @@ def compose_uri(fragment):
 
 def get_generic_data(dataset_id, entity_id, **kwargs):
     entity = get_entity(dataset_id, entity_id, **kwargs)
-    return json.loads(json.loads(entity.content)[0]['attrs'][0]['str_value'])
+    attrs = json.loads(entity.content)[0]['attrs']
```
clawconduce: `_kind` is an attr now & may be the first one

api.py
```diff
@@ -352,40 +356,28 @@ def create_dataset(dataset_name, **kwargs):
...
-        "timestamp_ms": timestamp,
-        "endtime_ms": timestamp,
-        "path": [location],
+        "time": timestamp,
+        "point": point,
         "removed": True,
     }
 
-    ingest_entities(dataset_id, {'entities': [remove_entity]}, **kwargs)
+    ingest_samples(dataset_id, [remove_entity], **kwargs)
```
clawconduce: I thought endless ents made more sense, but it is easier to stick with samples b/c of this remove call & the timestamp on that

konvulse: It does seem like `ingest_entities` makes more sense here.  Can you explain more about this choice?

api.py
konvulse: More robust for sure.

api.py
konvulse: I agree.  However this does keep the entities consistent.  Probably the best choice.  Making generic data go away will be a great triumph.

api.py
```diff
@@ -396,7 +388,7 @@ def convert_coordinates(point):
             'x': float(point['lon']),
             'y': float(point['lat'])
         }
-    elif 'x' and 'y' in sample['point']:
+    elif 'x' and 'y' in point:
```
konvulse: 😬

PR #66, opened 2017-11-29
Alex revisions

alex-conduce:
two outstanding issues that I wasn't able to indicate with edits alone: 

1. when I make a conduce-api request without adding an API key manually, I don't get a password prompt as the documentation seems to indicate. Instead I just get an Authorization required error basically. I'm making a ticket for this. 

2. When to use ingest_samples vs ingest_entities is still a bit confusing. 
   i. An entity is a list of samples right? so why can't I use api.ingest_samples(dataset_id, entity,...) ? 
   ii. What do I do if some of the records I am adding are samples (part of an existing entity already in the dataset) and some are entities (haven't been seen before)? Is this allowed?
   iii. What if I don't know whether something is an entity or a sample? There might be more updates to this id coming along but I'm not sure. Sometimes I don't know if this entity exists in Conduce already. If its the first time, I suppose its an entity, but if not, its a sample. Is this correct?

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ All CLI functionality is available through `conduce/api.py` and all API function
 As an example, you may list existing datasets with the following command:
 
 ```
-conduce-api list datasets --host=app.conduce.com --user=email@example.com --api-key=<your api key>
+conduce-api list datasets --host=app.conduce.com --user=email@example.com
 ```
 
 ## Setup
--- a/README.md
+++ b/README.md
@@ -51,7 +51,6 @@ The Conduce command line interface provides users with utilities for exercising 
 - Creating a dataset from a CSV or JSON file
 - Listing datasets and other resources
 
-```
 All CLI functionality is available through `conduce/api.py` and all API functionality should be usable with or without a `.conduceconfig` file. If you want to use the CLI without any setup you may fully specify your requests on the command line.
 
 As an example, you may list existing datasets with the following command:
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ import conduce
 ```
 Then you may use API functions, like `create_dataset` as follows:
 ```
-conduce.api.create_dataset('my-dataset-name', host='app.condcuce.com', user='email@example.com')
+conduce.api.create_dataset('my-dataset-name', host='app.conduce.com', user='email@example.com')
 ```
 However, this requires that the user authenticate with their password when the script is executed.  To avoid this, provide an `api-key` argument instead of `user`.  If you would like to learn more about API key creation, take a look at [API Key Creation](https://conduce-conduce-python-api.readthedocs-hosted.com/en/latest/api-key-creation.html).
 
@@ -51,6 +51,7 @@ The Conduce command line interface provides users with utilities for exercising 
 - Creating a dataset from a CSV or JSON file
 - Listing datasets and other resources
 
+```
 All CLI functionality is available through `conduce/api.py` and all API functionality should be usable with or without a `.conduceconfig` file. If you want to use the CLI without any setup you may fully specify your requests on the command line.
 
 As an example, you may list existing datasets with the following command:
@@ -87,7 +88,7 @@ or create an API key by adding the `--new` flag
 conduce-api config set api-key --new --user=valid_user --host=valid_host
 ```
 
-If you provide an API key for your default user and host you will never have to enter credentials to use that account as long as the API key is valid.
+If you configure an API key for your default user and host you will never have to enter credentials to use that account as long as the API key is valid.
 
 
 ## Conduce Configuration
--- a/README.md
+++ b/README.md
@@ -44,17 +44,24 @@ Other API features are documented in the [API Reference](https://conduce-conduce
 
 ## Getting started
 
-The Conduce command line interface provides users with utilities for exercising common API commands, and textual feedback for inspecting and debugging.  If you want to use the CLI without any setup you may fully specify your requests on the command line.
+The Conduce command line interface provides users with utilities for exercising common API commands, and textual feedback for inspecting and debugging. Features include:
+
+- Creating API keys
+- Adding and removing datasets
+- Creating a dataset from a CSV or JSON file
+- Listing datasets and other resources
+
+All CLI functionality is available through `conduce/api.py` and all API functionality should be usable with or without a `.conduceconfig` file. If you want to use the CLI without any setup you may fully specify your requests on the command line.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -44,17 +44,25 @@ Other API features are documented in the [API Reference](https://conduce-conduce
...
+- Listing datasets and other resources
+
+```
+All CLI functionality is available through `conduce/api.py` and all API functionality should be usable with or without a `.conduceconfig` file. If you want to use the CLI without any setup you may fully specify your requests on the command line.
 
 As an example, you may list existing datasets with the following command:
 
 ```
-conduce-api list datasets --host=app.conduce.com --user=email@example.com
+conduce-api list datasets --host=app.conduce.com --user=email@example.com --api-key=<your api key>
```
konvulse: I think this needs to change back.  If the user thinks they need to specify an API key here that will slow them down.  Remember, the user doesn't know about protools and they've never done anything with Conduce, they don't have an API key.  Username/password authentication is all that is available.

README.md
```diff
@@ -80,18 +88,7 @@ or create an API key by adding the `--new` flag
...
-
-The CLI provides many high level functions.  These include:
-
-- Creating API keys
-- Adding and removing datasets
-- Creating a dataset from a CSV or JSON file
-- Listing datasets and other resources
-
-All CLI functionality is available through `conduce/api.py` and all API functionality should be usable with or without a `.conduceconfig` file.
+If you configure an API key for your default user and host you will never have to enter credentials to use that account as long as the API key is valid.
```
konvulse: We don't cover setting a default user in `.conduceconfig`.  Maybe this needs an update to.

doc/source/api-key-creation.rst
```diff
@@ -4,7 +4,7 @@
 API Key Creation
 ====================
 
-API keys provide a flexible mechanism for granting access to your account and are used to authenticate requests made using the Conduce REST API.  They can be used in place of username/password authentication.  You may create as many API keys as you need, and may wish to use different API keys for performing specific API operations.  API keys can also be used to grant other users temporary access to your account.
+API keys are used to authenticate requests made using the Conduce REST API.  They can be used in place of username/password authentication.  You may create as many API keys as you need, and you may wish to use different API keys for performing specific API operations.  API keys can also be used to grant other users temporary access to your account.
```
konvulse: I like the first edit, but the addition of "you" is not necessary.

konvulse: Actually, what prompted you to remove "provide a flexible mechanism for granting access to your account and "?

doc/source/data-ingest.rst
```diff
@@ -31,9 +31,10 @@ Datasets hold collections of related entities and provide the interface a user n
 
 :py:func:`api.create_dataset`::
 
-    world_cities_dataset_id = conduce.api.create_dataset('world cities', host='app.conduce.com', api-key='00000000-0000-0000-0000-000000000000')
+    world_cities_dataset = conduce.api.create_dataset('world cities', host='app.conduce.com', api-key='00000000-0000-0000-0000-000000000000')
+    world_cities_dataset_id = world_cities_dataset['id']
 
-The dataset ID returned is a UUID.  It will be passed as a function parameter to the :py:func:`api.ingest_entities` function.
+The dataset object retuned contains a dataset ID. This is a UUID which will be passed as a function parameter to the :py:func:`api.ingest_entities` function.
```
konvulse: Does this return an object or just the ID?

doc/source/api-key-creation.rst
alex-conduce: you're right. Is that comma right before that necessary? It sounded a bit strange to me

doc/source/data-ingest.rst
konvulse: You're right, good catch!

konvulse: We need to provide an example return value here.

README.md
alex-conduce: you're right. I made this back before I had the password prompt working

README.md
alex-conduce: The original line was: 

If you provide an API key for your default user and host you will never have to enter credentials to use that account as long as the API key is valid.

The wording was confusing as I wondered whether providing an API key with a flag for instance would make it so that I never again had to preset my username, password, or environment, or if providing meant something more (like configuring)

My assumption is that this is incorrect. I would have to provide the API key every time instead of the username, password, and env OR configure it permanently. Thus I wanted to use "configure".

doc/source/api-key-creation.rst
alex-conduce: It seemed like filler/fluff. Isn't that what any api key does? 

Though actually, 1 thing I have started to see as specific to our api keys is that they substitute for a user, their password, and a specific environment. Usually I think of an API key as just specific to a user(with a password).

So maybe that could be put back in as something like: API keys provide a flexible mechanism for granting access to your account on a specific conduce environment.

What do you think?

README.md
konvulse: Nevermind, I was mistaken.

README.md
```diff
@@ -44,7 +44,15 @@ Other API features are documented in the [API Reference](https://conduce-conduce
...
 
-The Conduce command line interface provides users with utilities for exercising common API commands, and textual feedback for inspecting and debugging.  If you want to use the CLI without any setup you may fully specify your requests on the command line.
+The Conduce command line interface provides users with utilities for exercising common API commands, and textual feedback for inspecting and debugging. Features include:
+
+- Creating API keys
+- Adding and removing datasets
+- Creating a dataset from a CSV or JSON file
+- Listing datasets and other resources
+
+```
```
konvulse: This needs to be removed.

konvulse:
Let's try to have this wrapped up and merged today.

doc/source/api-key-creation.rst
```diff
@@ -4,7 +4,7 @@
 API Key Creation
 ====================
 
-API keys are used to authenticate requests made using the Conduce REST API.  They can be used in place of username/password authentication.  You may create as many API keys as you need, and may wish to use different API keys for performing specific API operations.  API keys can also be used to grant other users temporary access to your account.
+API keys provide a flexible mechanism for granting access to your account. They are used to authenticate requests made to the Conduce REST API in place of username/password authentication. You may create as many API keys as you need, and may wish to use different API keys for performing specific API operations.  API keys can also be used to grant other users temporary access to your account.
```
konvulse: s/in place of/as an alternative to/

PR #67, opened 2017-12-01
Font change

alex-conduce:
3 things: 

1. Does this look good? screen shot 2017-12-01 at 1 15 49 pm I changed only the body to gotham. Should I also change the headers to gotham? 
2. This method of changing the font is for Sphinx 1.6 and above. It won't work for anyone who tries to build the documentation with 1.5 or lower
3. Gotham is a custom font. It has to be downloaded onto the builder's computer and paid for. If we want everyone to be able to build this anywhere, should we consider a google font?


konvulse:
I published this branch:
https://conduce-conduce-python-api.readthedocs-hosted.com/en/font-change/

so we can see what the deployed product will look like.

doc/source/conf.py
```diff
@@ -105,6 +104,8 @@
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
+def setup(app):
+    app.add_stylesheet('my-styles.css')
```
konvulse: might want to name this file `conduce-style.css`

conducemike:
In Safari, the fonts don't look like Gotham. I'm not sure if we care. They look correct in Chrome.

konvulse:
OBE #69

PR #73, opened 2017-12-08
SKW-2183

alex-conduce:


doc/source/conf.py
```diff
@@ -55,7 +55,7 @@
 master_doc = 'index'
 
 # General information about the project.
-project = u'Conduce Python API'
```
konvulse: I don't think we need both changes.  I'd like to see what it looks like without this change.

doc/source/_static/conduce-style.css
```diff
@@ -22,3 +22,7 @@
     font-family: "Roboto Mono";
     src: url("fonts/RobotoMono/RobotoMono-Medium.ttf") format("truetype");
    }
+
+.logo {
+     margin-bottom: 10px;
```
konvulse: I think we want to bump this up to at least `40px`

alex-conduce:
screen shot 2017-12-11 at 10 45 00 am


current appearance. I think it looks good

PR #75, opened 2017-12-20
Skw 2213

alex-conduce:
This process also made me make this: https://sb-mct.atlassian.net/browse/SKW-2278

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,8 @@ import conduce
 ```
 Then you may use API functions, like `create_dataset` as follows:
 ```
-conduce.api.create_dataset('my-dataset-name', host='app.conduce.com', user='email@example.com')
+conduce.api.create_dataset('my-dataset-name',
+     host='app.conduce.com', user='email@example.com')
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -56,7 +57,8 @@ All CLI functionality is available through `conduce/api.py` and all API function
 As an example, you may list existing datasets with the following command:
 
 ```
-conduce-api list datasets --host=app.conduce.com --user=email@example.com
+conduce-api list datasets --host=app.conduce.com
```
konvulse: Examples like this should be copy paste-able.  So we should either not include them in this change and allow them to scroll or escape the newlines with backslashes.  I think they are probably more readable on a single line.

doc/source/data-ingest.rst
```diff
@@ -39,7 +39,8 @@ The dataset object retuned contains a dataset ID such as '55611f65-c657-45b1-9d0
 If you want to ingest data into an existing dataset you will need to acquire its dataset ID.  Datasets are Conduce resources.  All resources have an ID and can be queried with the REST API.  The Conduce Python API provides a convenience method (:py:func:`api.list_datasets`) for listing datasets::
 
 
-    datasets = conduce.api.list_datasets(host='app.conduce.com', api-key='00000000-0000-0000-0000-000000000000')
+    datasets = conduce.api.list_datasets(host='app.conduce.com',
```
konvulse: Is there an extra newline here?

PR #77, opened 2018-01-05
Added command for editing text based resources

konvulse:
This change adds an edit command to `conduce-api` that automatically opens a resource in an editor and updates the resource after the user saves changes and exits the editor.

I've tested this with editing hoverbox configuration.  I'm not sure what happens if there is invalid syntax.  I'm hoping `json` covers that.

### DIFF
--- a/conduce.py
+++ b/conduce.py
@@ -5,6 +5,10 @@ import json
 import api
 import util
 import base64
+import tempfile
+import os
+import sys
+from subprocess import call
 
 
 def list_from_args(args):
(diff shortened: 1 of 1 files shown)


konvulse:
Please let me know if you have any ideas on how to improve the usability of this feature.

elliottconduce:
I was able to verify that this works by changing the `arrange_menu_expanded` UI state value of an orchestration from `false` to `true`. It should be noted that seeing the change required closing the browser tab with the open orchestration, opening a new tab and re-opening the edited orchestration -- but that seems expected.

I was also able to verify that "breaking" a resource in various ways ( **1** removing the `y` key-value pair from the `camera` field, **2** creating a typo in the `camera` field's key, and **3** replacing a float value with a string value) did not successfully edit the resource _(despite the text: "Updating modified resource"):_
```
[elliott@work conduce-python-api]$ conduce-api edit --api-key iQUujxdCSUK4K3_oldl7YG56-JbG403BifGA8ZJLTZg --id dc1cd706-dad7-4160-a386-36421ddcb241
Updating modified resource
400 Client Error: Bad Request for url: https://banksy-dev.conduce.com/conduce/api/v2/resources/dc1cd706-dad7-4160-a386-36421ddcb241
Could not read json request: cannot unmarshal input

[elliott@work conduce-python-api]$
``` 

^ This can be compared to a successful edit which looks like this:
```
[elliott@work conduce-python-api]$ conduce-api edit --api-key iQUujxdCSUK4K3_oldl7YG56-JbG403BifGA8ZJLTZg --id dc1cd706-dad7-4160-a386-36421ddcb241
Updating modified resource
[elliott@work conduce-python-api]$
```

It should also be noted that the React app appears to recover gracefully from a botched edit of the UI state JSON blob, in that it can still be saved (which fixes the JSON blob) even after displaying an error banner such as: `Unexpected token g in JSON at position 84`.
**However**, the same can not be said for the Angular app, which fails to display the map and does not correct on save. _This means that a botched edit of the UI state will break an orchestration in the Angular app_

Overall, this PR looks good to me. However, it might be worthwhile to have QA or other devs try to break stuff with it, just to ensure we don't end up supporting a command that could potentially severely corrupt an existing orchestration file.

konvulse:
It might be a good future enhancement to re-open the temp file if the edit fails.  There is also no good way to abort an edit after writing to the temp file.

Regarding resource corruption,  I don't think that is the responsibility of this feature.  The server should prevent changes to resources that would cause server side issues.  The frontend should handle corrupt resources gracefully.

conduce.py
```diff
@@ -338,10 +379,17 @@ def main():
...
     parser_config_find.add_argument('--regex', help='An expression to match datasets and query')
-    parser_config_find.add_argument('--content', help='Content to retreive: id,full,meta')
+    parser_config_find.add_argument('--content', help='Content to retrieve: id,full,meta')
     parser_config_find.add_argument('--decode', action='store_true', help='Decode base64 and JSON for full content requests')
     parser_config_find.set_defaults(func=find_resource)
 
+    parser_config_edit = subparsers.add_parser('edit',  parents=[api_cmd_parser], help='edit resources that match the given parameters')
+    parser_config_edit.add_argument('--type', help='Conduce resource type to edit')
+    parser_config_edit.add_argument('--name', help='The name of the dataset to query')
+    parser_config_edit.add_argument('--id', help='The ID of the dataset to query')
```
elliottconduce: The help messages should refer to `resource` instead of `dataset`.

konvulse: Good catch.  Thanks.

PR #78, opened 2018-01-10
SKW-2267: Add json to csv util

clawconduce:


konvulse:
Should we have just used pandas?
http://pandas.pydata.org/pandas-docs/stable/index.html

konvulse:
It seems like it would be nice to have a function that converts from JSON to CSV and a function that translates the headers.  Although I suppose something like this is nice if you want to be selective about which fields you include in the CSV.

clawconduce:
This does assume flat dicts within a list.  I could see a similar function that has keys with values that are lists, and for both those situations, the values inside the dicts/list could be objects.

I wanted to avoid pandas b/c it's heavy mainly.  I am also not familiar with it, and the use case I had was decently trivial to solve using the csv lib & I have control over what happens with missing/null values.

Is this function worth generalizing now?  There are a lot of possibilities that could be covered, what is in this PR works (well, I have to add some changes to set delimiters and quoting), and I think with the new name `dicts_to_csv` is specific enough to say this converts a list of dicts to a csv.

PR #85, opened 2018-01-23
Added image upload command

konvulse:
Added a command for uploading images.  This is a convenience for creating asset resources for image mime types.

### DIFF
--- a/asset.py
+++ b/asset.py
@@ -1,0 +1,41 @@
+import mimetypes
+import api
+import util
+
+
+def initialize_binary_asset(path, **kwargs):
+    with open(util.walk_up_find(path, 'rb')) as image_stream:
+        mime = mimetypes.guess_type(path)[0]
+        image = image_stream.read()
+
+        name = kwargs.get('name', path.replace('/', '-'))
+        if name is None:
+            name = path.replace('/', '-').strip('-').lower()
+        print path, name
+        del kwargs['name']
+        asset_id = initialize_asset(name, image, mime, **kwargs)
+        return asset_id
+
+
+def initialize_asset(name, data, mime_type, **kwargs):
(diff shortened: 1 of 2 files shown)


konvulse:
I used this to upload an alternate substrate image for Dabrowa.  Then I applied it to the substrate with the edit command.

asset.py
```diff
@@ -0,0 +1,41 @@
...
+        del kwargs['name']
+        asset_id = initialize_asset(name, image, mime, **kwargs)
+        return asset_id
+
+
+def initialize_asset(name, data, mime_type, **kwargs):
+    asset_id = kwargs.get('id')
+    found = []
+
+    if asset_id is not None:
```
clawconduce: nitpick, best to call `modify_asset_content` in only 1 place in case there are future updates to the function:

```
if asset_id is not None:
    found = [asset_id]
elif kwargs.get('modify'):
    found = api.find_asset(name=name, **kwargs)
...
```

asset.py
```diff
@@ -0,0 +1,41 @@
+import mimetypes
+import api
+import util
+
+
+def initialize_binary_asset(path, **kwargs):
+    with open(util.walk_up_find(path, 'rb')) as image_stream:
```
clawconduce: why limit this to images?  Couldn't this also create json assets?

konvulse: This function is specifically for reading binary files.

We could add another convenience method for reading text files, but that doesn't seem as useful.  `initialize_asset` can be called directly, or the user can call `api.create_asset`.

asset.py
konvulse: This code is admittedly not very clean and could stand a redesign.  Part of the problem is that creation of the asset object is tied up in the the function to post it.

PR #86, opened 2018-01-23
Added command for editing entities from a dataset

konvulse:
This command enables a user to edit a single instance of an entity.  The
user specifies the dataset ID and the entity ID with an optional date
string.  The entity is opened in an editor for modification, after which
the modified entity is updated in the dataset.

### DIFF
--- a/conduce.py
+++ b/conduce.py
@@ -105,9 +105,10 @@ def edit_entity(args):
             if entity['timestamp_ms'] == util.string_to_timestamp_ms(args.date):
                 subject = entity
 
-    edited_entity = open_in_editor(json.dumps(subject, indent=2))
-    if edited_entity != str(subject):
-        return api.modify_entity(dataset_id, json.loads(edited_entity), **vars(args))
+    if subject is not None:
+        edited_entity = open_in_editor(json.dumps(subject, indent=2))
+        if edited_entity != str(subject):
+            return api.modify_entity(dataset_id, json.loads(edited_entity), **vars(args))
 
     return "No entity found"
 
--- a/conduce.py
+++ b/conduce.py
@@ -42,6 +42,17 @@ def find_resource(args):
     return resources
 
 
+def open_in_editor(content):
+    EDITOR = os.environ.get('EDITOR', 'vim')
+    with tempfile.NamedTemporaryFile(suffix='.tmp') as resource_file:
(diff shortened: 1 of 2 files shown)


konvulse:
I used this to modify the MJLR timezone in the sites dataset without having to run build_dhl.py.

conduce.py
```diff
@@ -42,6 +42,17 @@ def find_resource(args):
...
 
 
+def open_in_editor(content):
+    EDITOR = os.environ.get('EDITOR', 'vim')
+    with tempfile.NamedTemporaryFile(suffix='.tmp') as resource_file:
+        resource_file.write(str(content))
+        resource_file.flush()
+        call([EDITOR, resource_file.name])
+
+        with open(resource_file.name, 'r') as edited_resource_file:
```
clawconduce: Git has this thing with interactive rebases:

`# if you remove everything, the rebase will be aborted.`

It would be great to do something like that here, where there is a documented way to do nothing

conduce.py
```diff
@@ -79,6 +90,28 @@ def edit_resource(args):
...
+
+    subject = None
+    if args.date is None:
+        subject = entity_history[-1]
+    else:
+        for entity in entity_history:
+            if entity['timestamp_ms'] == util.string_to_timestamp_ms(args.date):
+                subject = entity
+
+    edited_entity = open_in_editor(json.dumps(subject, indent=2))
```
clawconduce: if the user types in the wrong date, `subject` could be `None`, so there should be a check, `if subject is None: raise ValueError("{} is an invalid date, use one of the following: {}"...)`

conduce.py
konvulse: I thought the same thing.  Right now, if the edited version is string equivalent to the original, nothing will happen.

I supposed we could also add a feature to exit gracefully if the user deletes the entire contents and attempts to commit those changes.

I'm not sure how we would document it though, outside of putting it in argparse.

conduce.py
konvulse: Admittedly I did not test the timestamp feature.  It doesn't seem all that useful.

conduce.py
konvulse: Also, deleting the contents of the file and saving it should have the desired effect, although I suspect the empty string will be sent to Conduce before failing.

PR #91, opened 2018-03-06
Added command for creating resources

konvulse:
Also updated the package description.

This is a basic command for creating resources.  It assumes the user has a fully defined content file.  It does no content validation.

I tested it by creating a lens template with `samples/world-cities-lens.json`.  I was then able to edit the resource to add the dataset ID and make it valid.

### DIFF
--- a/conduce.py
+++ b/conduce.py
@@ -10,6 +10,7 @@ import os
 import sys
 from subprocess import call
 import asset
+import mimetypes
 
 
 def list_from_args(args):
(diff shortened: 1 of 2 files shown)


conduce.py
```diff
@@ -22,6 +23,25 @@ def list_from_args(args):
     return api.list_resources(object_to_list.upper().rstrip('S'), **vars(args))
 
 
+def create_resource(args):
+    resource_type = args.type.upper()
+    resource_name = args.name
+    content_path = args.content
+
+    del vars(args)['name']
```
clawconduce: does this work?  I'd assume that vars(args) created a new variable, and name was deleted from that.  I think this should be something like `kwargs = vars(args); del kwargs['name'] ...`

conduce.py
```diff
@@ -22,6 +23,25 @@ def list_from_args(args):
...
+    del vars(args)['type']
+    del vars(args)['content']
+
+    with open(content_path, 'rb') as content_stream:
+        if resource_type.lower() == 'asset':
```
clawconduce: nitpick, but if you're in here, it seems weird to determine the mime type right after opening the file, even though it's not using anything from the file content.  I'd move this before the `with`

conduce.py
clawconduce: looks like it works with python 2 at least

konvulse: It works.  I use that pattern throughout.

conduce.py
konvulse: I thought about that.  The reasoning is that there is no point in determining the mime type if the file will not open.

PR #92, opened 2018-03-06
Added example script that creates world cities lens

konvulse:
Create a dataset, ingest the world cities data, create a lens template that points to the data, and set public permissions to view the data.

This may be useful for populating deploys.  This works with an API key or one-time password authentication.

### DIFF
--- a/samples/create_world_cities.py
+++ b/samples/create_world_cities.py
@@ -1,0 +1,43 @@
+import sys
+sys.path.append('..')
+
+import api
+import util
+import acl
+import json
+
+
+def create_world_cities(args):
+    kwargs = vars(args)
+    print "Creating dataset"
+    dataset_meta = api.create_dataset('world-cities-data', **kwargs)
+    print "Ingesting data from CSV"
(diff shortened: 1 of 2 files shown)


samples/create_world_cities.py
```diff
@@ -0,0 +1,43 @@
...
+import acl
+import json
+
+
+def create_world_cities(args):
+    kwargs = vars(args)
+    print "Creating dataset"
+    dataset_meta = api.create_dataset('world-cities-data', **kwargs)
+    print "Ingesting data from CSV"
+    util.ingest_file(dataset_meta['id'], csv='simplemaps-worldcities-basic.csv', kind='city', answer_yes=True, generate_ids=True, **kwargs)
```
konvulse: A better sample would not use util, but there are other commands used here that are not yet covered by documentation.  At some point this should be replaced with a local function that builds the entities.

clawconduce: where does the csv come from?

konvulse:
We'll have to add a request for that.  I was working within the existing structure.  The documentation will also have to change to accomodate.

konvulse:
New task:
https://sb-mct.atlassian.net/browse/SKW-2654

PR #94, opened 2018-03-19
SKW-2578: Expose the create-account api to the Conduce Python CLI

conducemike:


conduce.py
```diff
@@ -560,6 +569,13 @@ def main():
     parser_account_exists.add_argument('email', help='The email address to check')
     parser_account_exists.set_defaults(func=account_exists)
 
+
+    parser_create_account = subparsers.add_parser('create-account', parents=[api_cmd_parser], help='Create a new user by name and email')
+    parser_create_account.add_argument('name', help='A friendly name for the user. \"First Last\" has been our convention')
```
konvulse: The help strings should be written for external users.  This seems to have notes  that are only applicable to our team.  I like:

"The user's name as it will be displayed in Conduce."

conduce.py
```diff
@@ -560,6 +569,13 @@ def main():
...
+    parser_create_account.add_argument('email', help='The email address for the new user')
```
konvulse: "The email address associated with the new user"

conducemike:
@konvulse I updated the PR to match your specified changes.

PR #97, opened 2018-04-04
Added heatmap lens

konvulse:
Added a heatmap lens for world cities data.

I tested this on stg-app where world cities had never been loaded.


mbrennek:
kw_visible
kw-clipped

This really helps demonstrate the clipping/culling problem I'm trying to fix.  By moving the map 2px down and putting the city of Key West offscreen you can see it removes that whole bump from the heat map that key west was generating.

konvulse:
For posterity:

screen shot 2018-04-04 at 2 19 10 pm

PR #98, opened 2018-04-05
Added dataset metadata cli

konvulse:
```python conduce.py get-dataset-metadata --regex=.*cmx --host=dhl-stg-app.conduce.com --user=dhl-dev@conduce.com | jq .[0].t_max```

This command will help write shell scripts for generating dataset sensu alerts.

It probably makes sense to create a `conduce-api dataset` command soon considering how many dataset commands there are.


konvulse:
messed up the parser

PR #100, opened 2018-04-06
Added copy resource command

konvulse:
Creates a new resource with a new name from an existing resource.

Also fixed more copy paste errors in parser names.

### DIFF
--- a/conduce.py
+++ b/conduce.py
@@ -42,6 +42,40 @@ def create_resource(args):
         return api.create_resource(resource_type, resource_name, content, mime_type, **vars(args))
 
 
+def copy_resource(args):
+    if args.type is not None:
+        resource_type = args.type
+
+        if resource_type.lower() == "lenses" or resource_type.lower() == "templates":
+            resource_type = "LENS_TEMPLATE"
+
+        args.type = resource_type.upper().rstrip('S')
+
+    resources = api.find_resource(content='full', **vars(args))
+
+    if len(resources) > 1:
+        print "Found multiple resources:"
+        print resources
+        print "Please select a single resource and try again."
+        return
+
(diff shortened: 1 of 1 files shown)


conduce.py
```diff
@@ -42,6 +42,40 @@ def create_resource(args):
...
+        print "Found multiple resources:"
+        print resources
+        print "Please select a single resource and try again."
+        return
+
+    if len(resources) == 0:
+        print "Resource not found, please update your query and try again."
+
+    resource_name = args.resource_name
+    del vars(args)['resource_name']
```
clawconduce: would it make more sense to create a kwargs with the data that should stay?

conduce.py
```diff
@@ -42,6 +42,40 @@ def create_resource(args):
...
+
+        if resource_type.lower() == "lenses" or resource_type.lower() == "templates":
+            resource_type = "LENS_TEMPLATE"
+
+        args.type = resource_type.upper().rstrip('S')
+
+    resources = api.find_resource(content='full', **vars(args))
+
+    if len(resources) > 1:
+        print "Found multiple resources:"
```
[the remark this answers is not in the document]
konvulse: My intention was to print the list of resources.  However it's not printing them in a readable way.  I'll submit an update.

conduce.py
konvulse: I think of it as arguments should be passed unless explicitly removed.  That way if a new argument is added the default state is that it is passed all the way through to the lowest level function.  Whitelisting arguments would "break the chain" so to speak and require that the user add the argument at the command line level and the copy_resource function.  I probably went overboard, removing arguments I didn't need to.  Typically I only remove those that conflict with other arguments.

PR #102, opened 2018-04-12
SKW-2883: more robust waiting

konvulse:
This changes adds more careful status checks while waiting for ingest jobs to complete.  After this change `wait_for_job` will not return on a 500 error.  It will retry.  The function should still return immediately on a 400 error.

### DIFF
--- a/api.py
+++ b/api.py
@@ -133,15 +133,19 @@ def wait_for_job(job_id, **kwargs):
     """
     while True:
         time.sleep(0.5)
-        response = make_get_request(job_id, **kwargs)
-        response.raise_for_status()
+        try:
+            response = make_get_request(job_id, **kwargs)
 
-        if response.ok:
-            msg = response.json()
-            if 'response' in msg:
+            if response.ok:
+                msg = response.json()
+                if 'response' in msg:
+                    return response
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code < 500:
                 return response
-        else:
-            return response
+            else:
+                print "Job status check failed:", e.response.reason
+                print "Will retry after sleep period."
 
 
 def compose_uri(fragment):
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -133,15 +133,19 @@ def wait_for_job(job_id, **kwargs):
...
-        response = make_get_request(job_id, **kwargs)
-        response.raise_for_status()
-
-        if response.ok:
-            msg = response.json()
-            if 'response' in msg:
+        try:
+            response = make_get_request(job_id, **kwargs)
+
+            if response.ok:
```
konvulse: Technically, I think this check is not necessary.  An exception is raised if the response is not OK.

api.py
```diff
@@ -133,15 +133,19 @@ def wait_for_job(job_id, **kwargs):
...
+                msg = response.json()
+                if 'response' in msg:
```
konvulse: @jrmConduce should more strict checking be done here?

api.py
```diff
@@ -133,15 +133,19 @@ def wait_for_job(job_id, **kwargs):
...
+                    return response
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code < 500:
```
jrmConduce: This will probably work in our case, since a stuck job will eventually time out and result in a 404, but should we be more robust here?  Perhaps breaking out after some reasonable length of time when we get nothing but 5XX errors?

api.py
```diff
@@ -663,15 +667,14 @@ def _ingest_entity_set(dataset_id, entity_set, **kwargs):
         raise ValueError('parameter entity_set is not an \'entities\' dict')
 
     if kwargs.get('debug'):
+        kwargs['debug'] = False
```
jrmConduce: What's up with this?

Is this just an optimization to lift this out of the loop?

It does have slightly different behavior than before.  If entity_set['entities'] is empty, this will run whereas it wouldn't have before.

api.py
jrmConduce: When a job is still in progress, the api will return a json object containing only a `progress` field.

When a job is complete, the response will either be json containing `progress`, `response`, and `result` or plain text containing an error.  On any error, the http response won't be 200.

So I think this part is ok

api.py
konvulse: My intention was that this was just cleanup/optimization.  Thanks for pointing out the change in behavior, but I don't think it is relevant to execution in either mode.

api.py
konvulse: I was wondering that too, but to what end.  I am hoping this is good for now.

Considering a DHL ingest script. If we time out on 500s we'll just go to the next ingest job and start getting 500s there, presumably.

api.py
konvulse: Thanks, I know we've been over that before.

PR #105, opened 2018-04-21
Add password argument to create-api-key

konvulse:
Enable api key creation from bash script

### DIFF
--- a/session.py
+++ b/session.py
@@ -14,9 +14,9 @@ def api_key_header(api_key):
 
 def get_session(host, email, password):
     api_key = config.get_api_key(email, host)
-    #If a password is provided, it should override the API key
     if api_key and password is None:
         return api_key_header(api_key)
+
     cookies = None
     cookie_file_dir = os.path.join(os.path.expanduser('~'), '.conduce', host, email)
     if not os.path.exists(cookie_file_dir):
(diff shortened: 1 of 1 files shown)


jrmConduce:
I'm not entirely sure what's going on with the output here.

The first time I ran it, I got this:
```
Session expired

host: lab-app.conduce.com
user: enterprise_test@conduce.com
"FZFWlsI0SdiGyob9BD0LgOgB8UctSEiEvBAyK2j2NtE"
```

Then I ran against a different user and got this
```

host: lab-app.conduce.com
user: jrm@conduce.com
"ElFpBQgQT9-TaPtG2zao1Dj-9JueR0_JqsPxeno7wvc"
```

Then I ran against enterprise_test again and got just the key.

I couldn't really understand why it sometimes included the username/environment and sometimes it didn't.  I have no idea why the first run included a "Session expired" message.  I have no idea why there is sometimes a blank line in the output.  (Note that I always specified --host and --user while running these tests).

I'm also not really a fan of quoting the api key, it adds a needless processing step to scripts.

It's certainly functional though.  And I realize that this was keeping the existing behavior of the command while providing a more convenient option, so adjusting the output may be beyond the scope of this change.  And as far as scripting goes, we can always just pipe through `tail` and `sed`, so I'm inclined to call this good unless you think it's worth cleaning up the output to make life easier on scripters.

konvulse:
That does look like a bug.  When the user is prompted for a password the host and user are displayed to help inform the user of the correct password to enter.  Those should not be displayed when the credentials have been provided already.

PR #114, opened 2018-10-09
Add Sample dataset for colocated cities

mbrennek:
Added a small (4 entity) dataset that contains co-located cities.

Added City 1-4 at the real location of Chicago to test joined hover datasets.


mbrennek:
That's fine.  I didn't know where it should go, if anywhere, but I thought I'd offer.

PR #116, opened 2018-12-13
SKW-3620: Wait for datasets to clear

konvulse:
The internal API now returns a response object instead of True
The external API now waits for jobs to complete.

### DIFF
--- a/api.py
+++ b/api.py
@@ -693,7 +693,7 @@ def _clear_dataset(dataset_id, **kwargs):
     response = make_post_request(
         None, 'datasets/clear/{}'.format(dataset_id), **kwargs)
 
-    return True
+    return response
 
 
 def clear_dataset(**kwargs):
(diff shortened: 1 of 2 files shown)


konvulse:
I tested this using the command line utility.  I cleared 1 dataset by ID, twice.  I cleared 4 datasets by name, once.

api.py
```diff
@@ -727,6 +728,11 @@ def clear_dataset(**kwargs):
         else:
             return_message = "No matching datasets found."
 
+    for response in responses:
+        if 'location' in response.headers:
+            job_id = response.headers['location']
+            response = wait_for_job(job_id, **kwargs)
```
mbrennek: just noting for posterity that this will take at a minimum `responses.count * 0.5` seconds to complete, because the first thing wait_for_job does is sleep for half a second before checking to see if the job is finished.

konvulse: Gross.  Thanks for pointing that out.  Could probably use a good exponential backoff strategy there.

PR #119, opened 2019-01-17
Don't raise exception on numbers

konvulse:
Invalid syntax caused type checks to fail in all cases.  It went unnoticed because the exception was caught and the attributes were parsed as strings.

This bug would impact production for any ingest daemon that calls `dict_to_entities`.


konvulse:
I ingested the world cities data to dataset aab15516-4771-4559-ba5a-133129c91602 in dev-app. I checked the metadata to see that the pop attribute was an integer.

PR #120, opened 2019-03-18
Add --no-verify option to ignore SSL certificate errors

mbrennek:
Self-signed SSL certificates fail verification.  This PR adds a
--noverify command line flag to the conduce api utility to skip SSL
certificate verification.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = python -msphinx
+SPHINXBUILD   = python2 -msphinx
 SPHINXPROJ    = ConducePythonAPI
 SOURCEDIR     = source
 BUILDDIR      = build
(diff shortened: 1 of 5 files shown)


mbrennek:
I updated the session api to also check the --noverify flag, and plumbed it through the login request.

doc/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = python2 -msphinx
+SPHINXBUILD   = python -msphinx
```
mbrennek: whoops, this shouldn't be here

session.py
```diff
@@ -77,11 +77,17 @@ def login(host, email, password):
...
     if email is None:
         email = "dhl-dev@conduce.com"
 
-    print get_session(args.host, email, password)
+    if args.noverify:
+        verify = False
+    else:
+        verify = True
+
+    print get_session(args.host, email, args.password, verify)
```
mbrennek: also passed `args.password` here instead of just `password`, as `password` is undefined in this context

doc/Makefile
mbrennek: change removed

konvulse:
I suggest changing the option to `--no-verify`

PR #123, opened 2019-05-01
ingest entitiy set

konvulse:
Allows a user to ingest a "raw" entity set json file.  This was useful for ingesting a bin-centers file from a script of which the output was an entity set JSON.


mbrennek:
we should probably bump the version with this too, right?

konvulse:
It's not really an API change, just a change to one of the utilities.  So not sure.

konvulse:
Definitely a change to the CLI.

PR #128, opened 2019-05-08
SKW-3961: futurize conduce api

konvulse:
Applies stage 1 and stage 2 as outlined in the quickstart guide here:

https://python-future.org/automatic_conversion.html

The only deviation is in some of the imports (see conduce.py for example).  Imports are to follow the convention `from . import config` but this convention produced errors in both Python 2 and Python 3. I think the project would need restructuring to make it work.

Hashtags #python3 #python-future.


konvulse:
Additional changes so far:
argparse subcommands
pickle binary file operations

konvulse:
Successfully ran create_world_cities.py

konvulse:
After these changes python conduce/conduce.py cannot be run directly in a dev env.  ./conduce.py has been added for that purpose.  conduce/conduce.py has been renamed conduce/cli.py.

konvulse:
It occurs to me that the instructions in the readme may not work anymore.  Specifically: 
```
pip install git+git://github.com/ConduceInc/conduce-python-api
```

conduce/ingest.py
```diff
@@ -0,0 +1,28 @@
+from . import api
+from . import util
+
+
+def get_dataset_id(dataset_name, **kwargs):
```
konvulse: I don't really care for this function in a file called `ingest.py`.  It should either move to `api.py` or this filename should change.  I like the latter, but I can't think of a good name at this time.

konvulse: I suppose this function could be removed, but it provides a nice interface.  If `find_resource` had an option to return only the first result and not an array this function would not be necessary.  `find_resource(..., first=True).get('id')` would be equivalent

konvulse:
Ready for thorough review.

README.md
```diff
@@ -1,64 +1,70 @@
...
-`api.py` is the majority of the implementation.  `conduce.py` is a command line utility for invoking the API.  It also serves as a representative implementation of the library.
+`conduce/api.py` is the majority of the implementation. `conduce/cli.py` is a command line utility for invoking the API. It also serves as a representative implementation of the library.
 
 # Prerequisites
 
 Conduce Python API requires that you have installed the following:
 
-- Python 2.7: https://www.python.org/downloads/
-- pip: https://pip.pypa.io/en/stable/installing/
+* Python 2.7: https://www.python.org/downloads/
```
jrmConduce: So is this expected to work in both python 2 and python 3 currently?

jrmConduce:
General question: I know that we typically use the latest from git in requirements.txt wherever this library is used, and I know there are some breaking changes here.  Are there any outstanding libraries that need to be updated before this change could be merged?

README.md
konvulse: Actually, yes.  It should work with both.  I overlooked the prerequisites.

konvulse:
> General question: I know that we typically use the latest from git in requirements.txt wherever this library is used, and I know there are some breaking changes here. Are there any outstanding libraries that need to be updated before this change could be merged?

Yes, there were changes made to the DHL repositories to stop using functions that were moved to ingest.py.  I'm not sure if they have been merged yet.

konvulse:
DHL PRs have been merged.  We should be good to go here.

PR #137, opened 2019-07-18
Enable user to set permissions on list of resources

konvulse:
Resource ID argument is now parsed as a comma separated list (no spaces) and permissions are applied to all resources (one or more) in the list.


konvulse:
This PR should be merged last.  It has the highest version number.

PR #139, opened 2019-07-19
Added rename resource command

konvulse:


mbrennek:
maybe return the http response instead? or the http status of the http response?

konvulse:
Rebased and pushed with changes.

Now the rename command prints nothing by default.  The `-v` option causes a find resource call on the ID that was renamed and prints its information.

konvulse:
Updated to use `op: add` as some resources do not have names.

PR #141, opened 2019-07-23
don't crash on unnamed resources

konvulse:
print empty string when no name exists and "None" when no ID exists (impossible).


konvulse:
Tested with `conduce-api find -1` on dhl-prd-app where there are resources without names.

PR #146, opened 2019-08-05
util cleanup

konvulse:
This change moves some utility functions out of api.py.  These functions convert samples (as defined in the documentation) to entity sets.  They are valuable outside the api package.

EDIT:  I also threw in a documentation fix for some syntax error in examples.

### DIFF
--- a/conduce/util.py
+++ b/conduce/util.py
@@ -375,7 +375,7 @@ def csv_to_entities(infile, **kwargs):
 
 def _convert_coordinates(point):
     if len(point) != 2:
-        raise KeyError('A point should only have two dimensions', point)
+        raise KeyError('A point should only have two dimensions.', point)
     if 'lat' and 'lon' in point:
         coord = {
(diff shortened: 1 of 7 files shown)


conduce/util.py
```diff
@@ -372,6 +373,142 @@ def csv_to_entities(infile, **kwargs):
...
+
+def parse_samples(samples):
+    for sample in samples:
+        if 'time' in sample:
+            sample['time'] = parser.parse(sample['time'])
+    return samples
+
+
+def json_to_samples(json_file):
+
```
mbrennek: nittiest nitpick, but there's an extra newline here

setup.py
```diff
@@ -7,7 +7,7 @@
 
 setup(
     name="conduce",
-    version="3.3.1",
+    version="3.4.1",
```
mbrennek: not 3.4.0?

PR #147, opened 2019-08-06
SKW-4167: post transaction methods and cli

konvulse:
This change adds post transaction methods to the Python API.

Updates documentation and tests.

Fixes bad code in json_to_samples.

Not changing the version number until all API methods have been updated. (Let me know if that's a bad idea.)

### DIFF
--- a/conduce/api.py
+++ b/conduce/api.py
@@ -626,6 +626,7 @@ def _post_transaction(dataset_id, entity_set, **kwargs):
     should be avoided in favor of :py:func:`insert_transaction` and :py:func:`append_transaction`.
 
     By default, the transaction will not be processed. Use the \`process\` kwarg to trigger automatic processing.
+
     Parameters
     ----------
     dataset_id : string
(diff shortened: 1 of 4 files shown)


konvulse:
retest this please

konvulse:
Only Conduce_Python_API_Freestyle is a real test.

konvulse:
If this gets approved, please let me merge.  It conflicts with the code style PR.

konvulse:
This PR has been rebased against charlie/pycodestyle

conduce/api.py
```diff
@@ -598,6 +598,95 @@ def ingest_entities(dataset_id, entity_list, **kwargs):
     return _ingest_entity_set(dataset_id, entity_set, **kwargs)
 
 
+def insert_transaction(dataset_id, entity_set, **kwargs):
```
konvulse: It just occurred to me that these names may be too vague for the python API.  Should `insert_transaction` become `insert_dataset_transaction` and so on?

mbrennek: I think `insert_transaction` is fine.  We don't use transaction in any context other than around a dataset, correct?  What else might we add that has the concept of a transaction that must be inserted/appended/posted?

konvulse: No idea, terms like transaction and backend just don't seem specific (self-documenting) enough.

konvulse: For the backend APIs I used `create_dataset_backend`, `add_elasticsearch_store`, and `process_transactions`

PR #156, opened 2019-08-07
SKW-4167: get transaction methods and cli

konvulse:
This adds the GET endpoint and a passthrough CLI, the only difference being the API default to returning only the last transaction instead of all transactions.


konvulse:
Builds off the post_transaction API.  We should consider whether defaulting min to -1 is a good idea or not.  Maybe the paging feature prevents what that default was designed for.

konvulse:
Github reported conflicts but this rebased cleanly, locally.

PR #158, opened 2019-08-08
SKW-4168: create dataset backends

konvulse:
Adds generic method to create a backend, only valid backend types are accepted.

Adds convenience methods for all valid backend types.

No CLI support in this PR.

### DIFF
--- a/conduce/api.py
+++ b/conduce/api.py
@@ -818,7 +818,7 @@ def add_tile_store(dataset_id, auto_process, **kwargs):
     return _create_dataset_backend(dataset_id, DatasetBackends.tile, auto_process, None, **kwargs)
 
 
-def add_capped_tile_store(dataset_id, auto_process, minimum_temporal_level, maximum_temporal_level, **kwargs):
+def add_capped_tile_store(dataset_id, auto_process, min_temporal_zoom_level, min_spatial_zoom_level, **kwargs):
     """
     Adds a capped tile store to the specified dataset.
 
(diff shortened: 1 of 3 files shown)


conduce/api.py
```diff
@@ -756,6 +778,128 @@ def delete_transactions(dataset_id, **kwargs):
...
+    auto_process : boolean
+        Configure the backend to auto process transactions. Auto processing is enabled by default.
+        Set to False to disable auto processing.
+    **kwargs : key-value
+        See :py:func:`make_post_request` for more kwargs.
+    """
+    return _create_dataset_backend(dataset_id, DatasetBackends.tile, auto_process, None, **kwargs)
+
+
+def add_capped_tile_store(dataset_id, auto_process, minimum_temoral_level, maximum_temporal_level, **kwargs):
```
mbrennek: `minimum_temoral` level should be `minimum_temporal_level`

mbrennek: `minimum_temporal_level` and `maximum_temporal_level` are not documented in the parameters section.

konvulse: Wondering if `maximum_temporal_zoom` would be more clear.  Maybe `max_temporal_zoom_level`.

konvulse:
Hmm, the capped tile store method doesn't seem to actually work.  Checking...

konvulse:
Wow, I totally screwed up the config parameters.  max_temporal_level should be min_spatial_level.

conduce/api.py
konvulse: Since I botched those parameters, I did wind up changing them to shorten to min/max and include the word zoom for clarity.

mbrennek:
Going to merge this one to make reviewing the next two easier.  The Makefile can be fixed in a new PR.

PR #160, opened 2019-08-12
get_transactions max overrides min

konvulse:
I think based on the documentation it was impossible to use the max parameter.  This change removes min if max is set and adds tests.


mbrennek:
Where are you seeing that we should provide min & max in the same request?

Reading the documentation for `GET /api/v2/data/{datasetId}/transactions`, it invalid to use `min` with `value` or `max` with `value`, but it doesn't say anything about `min` and `max` together.  And reading the code, I believe it should be possible to set both and make a valid request.

konvulse:
I think I misread the documentation today.

PR #161, opened 2019-08-13
Do pyenv the right way

konvulse:
This change uses the PYENV_VERSION environment variable instead of the `pyenv local` command to set the python version when creating virtual environments.  Once the virtual environments are created pyenv has done its job and is no longer required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+
 deps: setup-env deps-venv2 deps-venv3
 
 setup-env: venv2 venv3 
@@ -5,11 +6,19 @@ setup-env: venv2 venv3
 venv2:
 	mkdir -p venv2
 	pyenv install -s 2.7.16
+ifneq "$(shell PYENV_VERSION=2.7.16 python -V 2>&1)" "Python 2.7.16"
+	echo "pyenv not configured correctly, check your configuration"
+	exit 1
+endif
 	PYENV_VERSION=2.7.16 virtualenv venv2
 
 venv3:
 	mkdir -p venv3
 	pyenv install -s 3.7.4
+ifneq "$(shell PYENV_VERSION=3.7.4 python -V 2>&1)" "Python 3.7.4"
+	echo "pyenv not configured correctly, check your configuration"
+	exit 1
+endif
 	PYENV_VERSION=3.7.4 python -m venv venv3
 
 deps-venv2: venv2 requirements-dev.txt
--- a/Makefile
+++ b/Makefile
@@ -4,19 +4,13 @@ setup-env: venv2 venv3
 
 venv2:
 	mkdir -p venv2
-	cd venv2
 	pyenv install -s 2.7.16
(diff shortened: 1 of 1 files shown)


konvulse:
_I think_, the original Makefile should have worked for everyone if pyenv were configured correctly on the users system.  It didn't work correctly, but it just so happened that setting the python version on the root directory for the project should have caused the virtual environments to be configured using the version specified by .python_version.

So, while the Makefile didn't do what I intended, the effect should have been what I desired.  Using PYENV_VERSION is more correct.

Makefile
```diff
@@ -1,22 +1,25 @@
...
 
 venv3:
 	mkdir -p venv3
-	cd venv3
 	pyenv install -s 3.7.4
-	pyenv local 3.7.4
-	cd ..
-	python -m venv venv3 
+ifneq "$(shell PYENV_VERSION=3.7.4 python -V 2>&1)" "Python 3.7.4"
+	echo "pyenv not configured correctly, check your configuration"
```
mbrennek: I'm hitting this statement on both macOS 10.14 and macOS 10.15.  On both I had previously installed pyenv with `brew install pyenv`.

After digging around a bit in how to use pyenv, I was able to get it to work by changing the previous line to `ifneq "$(shell PYENV_VERSION=3.7.4 pyenv exec python -V 2>&1)" "Python 3.7.4"`, as well as changing lines 13 and 22 to include the `pyenv exec` statement as well.

konvulse:
Ah, I bet that bypasses the requirement to run pyenv init in your shell.

mbrennek:
> Ah, I bet that bypasses the requirement to run pyenv init in your shell.

I was unaware that running `pyenv init` was a requirement.  That would probably explain our disconnect over this.  I've never used pyenv before seeing it in this makefile.

konvulse:
> > Ah, I bet that bypasses the requirement to run pyenv init in your shell.
> 
> I was unaware that running `pyenv init` was a requirement. That would probably explain our disconnect over this. I've never used pyenv before seeing it in this makefile.

I like your way.  I'll make the update.

konvulse:
I incorporated the change @mbrennek suggested and discovered that the variables were being set in the first pass of the makefile.  That meant that the versions were being checked before being installed.  I added a rule to check the versions before installing the virtual environments and stopped use `shell` which seems to be run only in the first pass of the makefile.

Then I realized that the conditionals were being checked in the first pass as well, before the virtual environments are guaranteed to exist.  Changing the conditions to bash syntax resolved that issue.

Also added a check to make sure the python version installed in the virtual environments is the expected version.

konvulse:
I also made sure that the python 2.7.16 version was installed correctly.  This led me to the realization that I had to install virtualenv after installing python.  I checked this workflow by uninstalling 2.7.16 `pyenv uninstall 2.7.16` and running `make clean && make`.

PR #162, opened 2019-08-14
SKW-4168: backends cli

konvulse:
Functionally tested by adding backends and processing transactions in several environments.

### DIFF
--- a/conduce/api.py
+++ b/conduce/api.py
@@ -850,6 +850,30 @@ def list_dataset_backends(dataset_id, **kwargs):
     return make_get_request('/api/v2/data/{}/backends'.format(dataset_id), **kwargs)
 
 
+def set_default_backend(dataset_id, backend_id, **kwargs):
+    """
+    Make the specified backend the default for the specified dataset.
+
+    Parameters
+    ----------
+    dataset_id : string
+        The UUID that identifies the dataset to which the backend belongs.
+    backend_id : string
+        The UUID that identifies the dataset backend to update.
+    **kwargs : key-value
+        See :py:func:`make_patch_request` for more kwargs.
+
+    Returns
+    -------
+    requests.Response
+         On success, an asynchronous job ID.  Check the job status for progress.
+         (See :py:func:`wait_for_job`)
+    """
+
+    fragment = '/api/v2/data/{}/backends/{}'.format(dataset_id, backend_id)
+    return make_patch_request({}, fragment, parameters={'default': 1}, **kwargs)
+
+
 def process_transactions(dataset_id, backend_id, **kwargs):
     """
     Process transactions on a dataset backend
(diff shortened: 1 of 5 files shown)


konvulse:
Removed Makefile change to avoid conflicts

conduce/api.py
```diff
@@ -847,7 +849,31 @@ def list_dataset_backends(dataset_id, **kwargs):
     **kwargs : key-value
         See :py:func:`make_get_request` for more kwargs.
     """
-    return make_get_request('/api/v2/data/{}/backends'.format(dataset_id), **kwargs)
+    return json.loads(make_get_request('/api/v2/data/{}/backends'.format(dataset_id), **kwargs).content)
+
+
+def set_default_backend(dataset_id, backend_id, **kwargs):
```
konvulse: This new api method is in its own PR now.

conduce/api.py
```diff
@@ -748,8 +748,10 @@ def get_transactions(dataset_id, **kwargs):
     # HACK: get around bool parameter bug
     if not kwargs.get('count'):
         parameters.pop('count')
+    if kwargs.get('max'):
+        parameters.pop('min')
```
jrmConduce: Why?

It is perfectly legal and reasonable to request a range of transactions, which is why a user would set both max and min.

conduce/api.py
```diff
@@ -981,7 +1007,7 @@ def add_tile_store(dataset_id, auto_process, **kwargs):
     return _create_dataset_backend(dataset_id, DatasetBackends.tile, auto_process, None, **kwargs)
 
 
-def add_capped_tile_store(dataset_id, auto_process, min_temporal_zoom_level, min_spatial_zoom_level, **kwargs):
+def add_capped_tile_store(dataset_id, auto_process, min_spatial_zoom_level, min_temporal_zoom_level, **kwargs):
```
jrmConduce: No default values?

conduce/cli.py
```diff
@@ -316,6 +316,169 @@ def get_dataset_transactions(args):
...
+
+    for backend_id in backend_ids:
+        location = api.remove_dataset_backend(dataset_id, backend_id, **vars(args)).headers['location']
+        print('Removing {}: check status at {}'.format(backend_id, location))
+
+
+def list_dataset_backends(args):
+    dataset_id = args.dataset_id
+    del vars(args)['dataset_id']
+    backends = args.backend_ids
```
jrmConduce: It seems weird for a method that is supposed to list backends to optionally take a list of backends.

I get what you're going for, but it feels like that should be a separate method.

conduce/api.py
jrmConduce: never mind this, I see the default is set below

conduce/api.py
konvulse: Oops, that was not supposed to be included.  #160

conduce/cli.py
konvulse: yeah, and i suppose it only works with the `-v` option

konvulse: I'll add a story to address this

konvulse: SKW-4227

PR #165, opened 2019-08-14
backend status cli

konvulse:
Found this to be quite helpful: 
```
[18:29:14] conduce-mbp-kvl:conduce-python-api kvl$ ./conduce-api dataset backend-status 8b86dbe3-9398-4cbd-8834-b5798916c9a1  --host=app.conduce.com 
8b86dbe3-9398-4cbd-8834-b5798916c9a1:
  HistogramStore::e7f6add3-beb7-11e9-829b-3215067658a1: processed 3464 of 3464 transactions
  CappedTileStore::a90ec4bc-beb7-11e9-b4a0-a2fac95a8ece: processed 2926 of 3464 transactions
```

### DIFF
--- a/conduce/cli.py
+++ b/conduce/cli.py
@@ -479,6 +479,23 @@ def add_capped_tile_store(args):
     return api.add_capped_tile_store(dataset_id, auto_process, min_spatial, min_temporal, **vars(args))
 
 
+def get_backend_status(args):
+    dataset_ids = args.dataset_ids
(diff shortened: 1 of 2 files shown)


konvulse:
It's not quite right, I'm not sure the transactions field on a backend represents the same quantity as the count on a transaction log.  I'm not even sure it matches between backend types, but it's better than nothing.

I suppose making job status more helpful would be a good compliment to this.

PR #167, opened 2019-08-19
Remove min/max if value specified

konvulse:
min defaults to -1 in the CLI.  This change causes value to override min/max rather than print an error.

An alternative would be to only remove min if -1 and value specified.


mbrennek:
tests fail

konvulse:
Rebased.  Updated tests and added test cases.

PR #170, opened 2019-08-20
SKW-4232: wait for jobs more patiently

konvulse:
with tests

### DIFF
--- a/conduce/api.py
+++ b/conduce/api.py
@@ -203,8 +203,8 @@ def wait_for_job(job_id, **kwargs):
     """
     initial_status_countdown = 10
     server_error_count = 0
+
     while True:
-        time.sleep(0.5)
         try:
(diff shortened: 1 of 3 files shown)


conduce/api.py
```diff
@@ -229,6 +229,7 @@ def wait_for_job(job_id, **kwargs):
                 if kwargs.get('cli'):
                     print("Job status check failed for {}:".format(job_id), e.response.reason)
                     print("Will retry after sleep period.")
+                    time.sleep(0.5)
```
konvulse: I think this change is not right.  I want to move the sleep to after the execution of make_get_request.  So I guess I need to add another sleep at the end of the try block.

konvulse:
Superseded by #172

konvulse:
Superseded by #172

PR #172, opened 2019-08-21
wait for job timeout

konvulse:


conduce/api.py
```diff
@@ -222,13 +230,19 @@ def wait_for_job(job_id, **kwargs):
...
+            elif server_error_count < 119:
                 server_error_count += 1
                 if kwargs.get('cli'):
                     print("Job status check failed for {}:".format(job_id), e.response.reason)
                     print("Will retry after sleep period.")
+                time.sleep(0.5)
+            else:
+                raise e
+
+        timeout -= 1
```
konvulse: Maybe this should only decrement if no exceptions are raised.

PR #174, opened 2019-08-23
Added timeout arg to process transactions command

konvulse:
This is the CLI companion the the changes to wait_for_job.  Allows a user to set a custom timeout.


konvulse:
In a custom script I'm using the process transactions for wwln and mbta I modified the code that processes transactions one at a time like so:

```
    max_transaction = api.get_transactions(dataset_id, count=True, host=host)['count'] - 1
    for backend_id in backend_ids:
        min_transaction = api.get_dataset_backend_metadata(dataset_id, backend_id, host=host)['transactions']
        idx = 1
        process_times = [kwargs.get('timeout_seed', 600)]
        while min_transaction + idx < max_transaction:
            try:
                now = datetime.datetime.now()
                print("{}: Processing transaction {} of {} on {}:{}...".format(
                    datetime.datetime.now(), min_transaction + idx, max_transaction, dataset_id, backend_id))
                print("DEBUG: Will timeout in {}s".format(statistics.mean(process_times) * 2))
                response = api.process_transactions(dataset_id, backend_id, transaction=(min_transaction + idx), host=host)
                api.wait_for_job(response.headers['location'], timeout=(statistics.mean(process_times) * 2), host=host)
                then = now
                now = datetime.datetime.now()
                process_times.append((now - then).total_seconds())

                idx += 1
            except api.TimeoutError as e:
                print(e)
```

This change adjusts the timeout according to the "actual time" it takes to process each transaction.  I say "actual time," because it assumes the loop is instantaneous.  Sleep time is not reduced to account for processing time.  I consider this to be a good thing as it eliminates variance in request/response time.

PR #178, opened 2019-08-26
zoom level by length

konvulse:
Enable users to set the minimum zoom level by specifying the dimension of the minimum zoom level.


konvulse:
Should have thought of that.  Wondering if I could get by with one function since the distance values for zoom level zero are so close.

PR #182, opened 2019-09-05
SKW-4251: Migrate to k8s Jenkins

konvulse:
Update Jenkinsfile to work with new Jenkins server.

Checkout the status of this build on the new server here:
https://jenkins.mct.io/job/conduce-python-api/view/change-requests/job/PR-182/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,81 +2,96 @@
 
 properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '14', numToKeepStr: '10')), [$class: 'ScannerJobProperty', doNotScan: false], [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false], pipelineTriggers([])])
 
+pod_template = """
+kind: Pod
+metadata:
+  name: conduce-python-api-build
+spec:
+  containers:
+  - name: build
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -2,81 +2,96 @@
...
 properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '14', numToKeepStr: '10')), [$class: 'ScannerJobProperty', doNotScan: false], [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false], pipelineTriggers([])])
 
+pod_template = """
+kind: Pod
+metadata:
+  name: conduce-python-api-build
+spec:
+  containers:
+  - name: build
+    image: python:3.7
```
jrmConduce: Were you going to try to build both python 2 and 3?  All you'd need to do is add a second container here and copy/paste some of the build/test stages below inside an appropriate container block

konvulse: I just figured I'd get it back to baseline here.  That sounds pretty straightforward though, so maybe I'll take a crack at it after this gets merged.

konvulse: For context, I tried to update the job to use the repos makefile.  I did this because several of the commands in the job did not migrate and could be replaced with make commands. (Make file came after original jenkins job.)

PR #183, opened 2019-09-06
add rule to make coverage

konvulse:
Added this when I was working on the Jenkinsfile updates.  Then didn't use make for Jenkins.  Didn't want to lose this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,11 @@ test3:
 docs: venv3 
 	. ./venv3/bin/activate && make -C doc html
 
+coverage: venv3
+	./venv3/bin/coverage run --source conduce -m unittest discover -v test/ 
+	./venv3/bin/coverage report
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -74,10 +74,17 @@ test3:
 docs: venv3 
 	. ./venv3/bin/activate && make -C doc html
 
+coverage: venv3
+	./venv3/bin/coverage run --source conduce -m unittest discover -v test/
```
mbrennek: `venv3/bin/coverage` is not installed by the venv3 rule, which is the only one you've specified as a dependency, but it is installed by the deps-venv3 rule.  So should he previous line be `coverage: deps-venv3` instead?

mbrennek: or are you assuming the user has done that first?

konvulse: I think I just set the wrong dependency.

